PR #132, opened 2022-12-29
docs: add s3 configuration

killme2008:
Adds doc for s3  storage options.


apdong2022 (commented):
make sure that in the chart, for File, it's ok to leave key and type empty

xtang (approved):

PR #187, opened 2023-02-15
fix: change schema to db in script handler

shuiyisong:
This pr mainly fixes the script handler doc on changing from `schema` to `db` in param


xtang (approved):
LGTM

PR #202, opened 2023-02-22
docs: add ping and health api for influxdb

Fengys123:


docs/user-guide/supported-protocols/influxdb.md
```diff
@@ -52,7 +52,35 @@ SELECT * FROM influxdb.monitor;
 ```
 
 The `/influxdb/write` supports query params including:
+
 * `db` specify which db to write, `public` by default.
 * `precision`, precision of timestamps in the line protocol. Accepts `ns` (nanoseconds), `us`(microseconds), `ms` (milliseconds) and `s` (seconds), nanoseconds by default.
 
+## PING
+
+GreptimeDB also simply support InfluxDB `ping` and `health` API.
```
xtang, line 61: Additionally, GreptimeDB also provides support for the ping and health APIs of InfluxDB.

Fengys123:
Fixed, PTAL @xtang

xtang (approved):
LGTM

PR #222, opened 2023-03-04
docs: improve docker run command

morigs:
Update Docker installation instructions so that:

1. Image pulled automatically
2. All ports exposed
3. Mount named volume instead of `pwd`
4. Container has name
5. Container automatically removed after stop

### DIFF
--- a/docs/installation/download-pre-built.md
+++ b/docs/installation/download-pre-built.md
@@ -19,8 +19,7 @@ curl -L https://raw.githubusercontent.com/GreptimeTeam/greptimedb/develop/script
 ## Docker
 
 ```shell
-docker pull greptime/greptimedb
-docker run -p 4002:4002 -v "$(pwd):/tmp/greptimedb" greptime/greptimedb standalone start
+docker run -p 4000-4004:4000-4004 -p 4242:4242 -v "greptime-vol:/tmp/greptimedb" --name greptime --rm greptime/greptimedb standalone start
 ```
 
 ## Next Steps


xtang:
@morigs Thanks for your contribution! Looks good to me.
@sunng87 Please take a look.

xtang (approved):

morigs:
Forgot to mention important fact: I've tried to run `latest` from docker.io with both old command and my command. In both cases all calls to API return empty response (no headers, no body), so it effectively doesn't work. I don't think it's because of the command, but perhaps I use API incorrectly or the image is broken

sunng87 (approved):

sunng87:
LGTM and thank you for reporting the connection issue. @morigs are you using linux as docker host? On linux, docker uses bridge network by default and greptimedb by default listens to 127.0.0.1, so it's not possible to connect via "curl http://127.0.0.1:4000". To fix that issue, linux user will need to add additional arguments, for example, to expose http and mysql protocol to host:

```
docker run -p 4000-4004:4000-4004 -p 4242:4242 -v "greptime-vol:/tmp/greptimedb" --name greptime --rm greptime/greptimedb standalone start --http-addr 0.0.0.0:4000 --mysql-addr 0.0.0.0:4002
```

I think we can add this to our docs as well

Wenjie0329 (approved):

morigs:
@sunng87 Thanks. I'm on mac, but you're right, servers listen on `127.0.0.1` by default, changing host to `0.0.0.0` helps.
Perhaps `0.0.0.0` should be default on all platforms?

sunng87:
@morigs Thank for your suggestion. I have initialized a discussion to see if 0.0.0.0 works for other developers. I think the only concern is about the security by default.

PR #256, opened 2023-03-16
refactor: refine download-pre-built.md

zyy17:


sunng87 (approved):
LGTM

sunng87:
We still need to address the default listen address issue with docker

docs/installation/download-pre-built.md
```diff
@@ -1,31 +1,54 @@
...
+-p 4242:4242 -v "greptime-vol:/tmp/greptimedb" \
+--name greptime --rm \
+greptime/greptimedb standalone start
 ```
 
+If you want to use another version of the GreptimeDB image, you can download it from our [GreptimeDB Dockerhub](https://hub.docker.com/r/greptime/greptimedb).
+
+**Note**:
+
+If the Docker version you are using is lower than [v23.0](https://docs.docker.com/engine/release-notes/23.0/), then you will encounter the problem of insufficient permissions when starting the above command (a [bug](https://github.com/moby/moby/pull/42681) of the old version of Docker Engine).
```
xtang, line 36:
```suggestion
If you are using a Docker version lower than [v23.0](https://docs.docker.com/engine/release-notes/23.0/), you may experience problems with insufficient permissions when trying to run the command above, due to a [bug](https://github.com/moby/moby/pull/42681) in the older version of Docker Engine.
```

xtang (approved):

PR #284, opened 2023-03-29
docs: update install cargo-nextest command

WenyXu:
See also: https://nexte.st/book/installing-from-source.html


xtang (approved):
LGTM from the wording.
@waynexia Could you take a look at the technical part?

waynexia (approved):

PR #425, opened 2023-06-17
feat: improve influxdb line document

NiwakaDev:
In `Table Management`, there is an example of creating a `monitor` table. In the following `Write Data/SQL` section, there is an example of inserting data into the `monitor` table. On the other hand, in `Write Data/InfluxDB Line` section, there is an example of inserting data into the `system_metrics` table. I guess that users should read `Write Data/InfluxDB Line` after they read `Table Management`. I think this might confuse them. 

So, this PR replaces `system_metrics` with `monitor` so that users can immediately execute a insert command after reading `Table Management`.

In addition, this PR appends a `InfluxdDB Line protocol` reference in the `Write Data/InfluxDB Line` section so that users could easily understand about the `InfluxdDB Line protocol`.

image

### DIFF
--- a/docs/user-guide/write-data/influxdb-line.md
+++ b/docs/user-guide/write-data/influxdb-line.md
@@ -7,15 +7,18 @@ GreptimeDB supports HTTP InfluxDB Line protocol. You can write data via `/influx
 ```shell
 curl -i -XPOST "http://localhost:4000/v1/influxdb/write?db=public&precision=ms" \
 --data-binary \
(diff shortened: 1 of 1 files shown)


xtang:
Thank you for pointing that out. You are correct, it's better to maintain consistency with the table name in the user guide.

xtang (approved):

xtang:
@nicecui please take a look.

nicecui (approved):
LGTM

PR #426, opened 2023-06-18
fix: prometheus basic_auth config

killme2008:
The valid `basic_auth` config is:

```
remote_write:
- url: http://localhost:5080/api/default/prometheus/api/v1/write
  basic_auth:
    username: greptime_user
    password: greptime_password
```

But the doc has an unnessary `-` before `basic_auth`, making the whole configuration invalid.

And the database authentication is disabled by default, so I think the `basic_auth` should be commented by default too.


nicecui (approved):
👍

xtang (approved):

PR #445, opened 2023-06-27
chore: change overview names in navigation

nicecui:


docs/getting-started/overview.md
```diff
@@ -1,4 +1,4 @@
-# Getting Started
+# Overview
```
sunng87, line 1: Since getting start has only one page, can we use this page directly?

sunng87:
Let's get this merged and leave discussion for next patch.

sunng87 (approved):

docs/getting-started/overview.md
nicecui: I'm not sure, maybe @alili has great ideas

PR #497, opened 2023-07-22
feat: update getting-started.md

xie-zheng:
- Purpose of this commit:
 add a single line of bash script to easily setup the nightly toolchain(required in 'rust-toochain.toml' in main project)

This is my first tiny commit to greptime community, 
looking forward to comments and  suggestions.

### DIFF
--- a/docs/developer-guide/getting-started.md
+++ b/docs/developer-guide/getting-started.md
@@ -28,6 +28,8 @@ Start GreptimeDB standalone instance in just a few commands!
 ```shell
 git clone https://github.com/GreptimeTeam/greptimedb.git
 cd greptimedb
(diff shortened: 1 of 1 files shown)


xtang:
@xie-zheng Thanks for contributing! Will take a look at it soon.

sunng87:
Hi @xie-zheng , thank you for your suggestion. But if you are using rustup to manage cargo binary, it will resolve this toolchain.toml when you run any cargo commands within the project. There is no need to install the toolchain manually. 

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

xie-zheng:
@sunng87 , Thanks for the reply. Guess the previous building failure is due to network problem, which failed the installation of nightly toolchain...

PR #500, opened 2023-07-25
feat: cookbook for commonly used cases

waynexia:
Add cookbook for commonly used cases including
- Query latest 5 minutes of data
- Group by time interval
- Cast number literal to timestamp
- Cast string literal to timestamp
- Extract the day of the year from timestamp


nicecui (approved):
👍

xtang (approved):

PR #521, opened 2023-08-09
docs: fix error links

QuakeWang:
This [link](https://www.docker.com/getting-started/) for getting started with docker has moved to https://www.docker.com/get-started/ .

Then some links in the document use relative paths, but cannot be accessed normally in GitHub, so add '...` prefix.

### DIFF
--- a/docs/getting-started/overview.md
+++ b/docs/getting-started/overview.md
@@ -3,9 +3,10 @@
 Begin to explore GreptimeDB's powerful core features.
 
 ## Install GreptimeDB
+
 You can try out GreptimeDB with our test builds released on [Download page](https://greptime.com/download).
 
-We use the simplest configuration for you to get started. For a comprehensive list of configurations available in GreptimeDB, see the [configuration documentation](../user-guide/operations/configuration.md).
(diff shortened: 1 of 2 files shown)


nicecui (commented):
Thank you very much for your contribution! Your help ensures that the Docker link will be correct.

However, it may not be necessary to make changes to relative paths. We use paths like `/user-guide` or `/reference` to reduce errors in internal hierarchical documentation. For example, it is easy to make mistakes when writing links to `reference` at `/user-guide/clients/sdk/go.md`, that would be `../../../reference/xxx.md`. In contrast, we can write `/reference/xxx.md`, which is simpler and less prone to errors.

We hope that all links will be recognized correctly on GitHub. However, our main priority is to publish the repository to https://docs.greptime.com without errors, with minimal need for considering link corrections. To achieve this, we require that all documents follow the same specification. Therefore, links to resource root directories in all documents should be written as `/reference`, `/user-guide`, and so on.

Thank you for your understanding and contribution! Expecting the modified PR.

QuakeWang:
> 

OK, got it. I will update it later : )

QuakeWang (commented):
@nicecui PTAL

nicecui (approved):
👍

xtang (approved):

PR #522, opened 2023-08-10
refactor: move cookbook to user-guide and reference

nicecui:
Move the cookbook to `/reference/sql/functions` and `user-guide/query-data/sql.md` for the following reasons:

- I think that the cookbook existed due to the lack of examples in querying data and the reduction of documentation on SQL functions.
- Documents would prefer to link to `reference` or `query-data` instead of linking to the cookbook. For example, linking to `reference/sql/functions.md` from `user-guide`, dashboard console, etc.
- Writing more and more examples in the cookbook over time is not a good idea because it requires organizing the content. I think the structure in the end would be the same as write-data or query-data.

Therefore, I moved the content to the expected place and removed the cookbook.


docs/reference/sql/functions.md
```diff
@@ -11,5 +15,140 @@ functions in DataFusion. These functions include:
...
+arrow_cast(expression, datatype)
+```
+
+Where the `datatype` can be any valid Arrow data type in this [list](https://arrow.apache.org/datafusion/user-guide/sql/data_types.html). The four timestamp types are:
+- Timestamp(Second, None)
+- Timestamp(Millisecond, None)
+- Timestamp(Microsecond, None)
+- Timestamp(Nanosecond, None)
+
+(Notice that the `None` means the timestamp is timezone naive)
```
sunng87, line 32: Our http API output them as 

- Timestamp
- TimestampMillisecond
- TimestampMicrosecond
- TimestampNanosecond

I think we can use these names here avoid explaining `None`

docs/reference/sql/functions.md
```diff
@@ -11,5 +15,140 @@ functions in DataFusion. These functions include:
...
+Please refer to [API documentation](https://greptimedb.rs/script/python/rspython/builtins/greptime_builtin/index.html#functions)
+
+## Time and Date
+
+### `date_trunc`
+
+`date_trunc` function follows the same API with PostgreSQL's [`date_trunc`](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC). It's illustrated as:
+
+```sql
+date_trunc(field, source [, time_zone ])
```
sunng87, line 45: rename `field` to a more meaningful name?

docs/reference/sql/functions.md
nicecui: No, that's the function input, not an API output.

`arrow_cast` function doesn't support expression like `select * from monitor where ts > arrow_cast(1690252336408, 'TimestampMillisecond');`

The error would be `Error during planning: Unsupported type 'TimestampMillisecond'. Must be a supported arrow type name such as 'Int32' or 'Timestamp(Nanosecond, None)'. Error unrecognized word: TimestampMillisecond`

docs/reference/sql/functions.md
nicecui: In the PostgreSQL's doc:

> `field` selects to which precision to truncate the input value.

maybe we could change `field` to `precision`

nicecui:
@sunng87 PTAL

sunng87 (approved):

docs/reference/sql/functions.md
```diff
@@ -11,5 +15,140 @@ functions in DataFusion. These functions include:
...
+- month
+- quarter
+- year
+- decade
+- century
+- millennium
+
+### INTERVAL
+
+The interval data type allows you to store and manipulate a period of time in years, months, days, hours etc. It's illustrated as:
```
xtang, line 65:
```suggestion
The Interval data type allows you to store and manipulate a period of time in years, months, days, hours etc. It's illustrated as:
```

xtang (commented):
Almost LGTM, and PTAL at the comments

xtang (approved):

PR #559, opened 2023-09-06
docs: add relationship between tags and primary keys

NiwakaDev:
Right now `greptimedb` doesn't support specifying tags when creating a table. So we need to specify primary keys for tags.

If you want to know more detail, please see https://github.com/GreptimeTeam/greptimedb/issues/2183.


NiwakaDev:
It seems like we've already written.

PR #614, opened 2023-10-11
docs: update external table related documents

zhongzc:
## What's Changed in this PR

External table related parts of documents for 0.4 are updated.

## Checklist


xtang (approved):

docs/en/v0.4/user-guide/query-external-data.md
```diff
@@ -97,14 +102,18 @@ SELECT * FROM yellow_tripdata LIMIT 5;
...
+|        2 | 2022-02-01 00:06:48  | 2022-02-01 00:33:07   |               1 |         12.16 |          1 | N                  |          138 |          125 |            1 |        35.5 |   0.5 |     0.5 |       8.11 |            0 |                   0.3 |        48.66 |                  2.5 |        1.25 | 1970-01-01 00:00:00 |
++----------+----------------------+-----------------------+-----------------+---------------+------------+--------------------+--------------+--------------+--------------+-------------+-------+---------+------------+--------------+-----------------------+--------------+----------------------+-------------+---------------------+
 5 rows in set (0.11 sec)
 ```
+
+:::tip Note
+The query result includes the value of the `greptime_timestamp` column, although it does not exist in the original file. All these column values are `1970-01-01 00:00:00+0000`, because when we create an external table, the `greptime_timestamp` column is automatically added with a default value of `0`, representing `1970-01-01 00:00:00+0000`. You can find more details in the [create](../reference/sql/create.md#create-external-table) document.
```
xtang, line 118: There are two "default values": `0` and `1970-01-01 00:00:00+0000`, and I'm concerned that this might confuse new users. Is `1970-01-01 00:00:00+0000` merely a display value?

zhongzc: The `0` here means that this is a zero value. You reminded me that there is no need to mention it. I will change it.

xtang (approved):

xtang:
@nicecui PTAL

nicecui (approved):

PR #683, opened 2023-11-09
fix: add doc version to the links on console

nicecui:
## What's Changed in this PR

add doc version to the links on console

## Checklist


xtang (approved):

sunng87:
Are we going to fix a bug or something? I doubt if we really need this: using default link would be OK since cloud will always be latest. Otherwise we will need to update those links on each rotate

PR #686, opened 2023-11-13
docs: `quick-start` of greptimeDB

nicecui:
## What's Changed in this PR

The old Getting Started

```
- Getting Started
  - Try Out GreptimeDB
  - Try Out GreptimeDB Cluster
 ```

The new Getting Started

```
- Getting-Started:
  - Installation:
    - greptimedb-standalone
    - greptimedb-cluster
  - Quick-Start:
    - overview
    - mysql
    - influxdb
    - go
    - java
    - python
    - node
    - vector
    - prometheus
```

## Checklist


Aganivi:
> I'm afraid we need auto redirect feature for this @alili @Aganivi

Could you describe it in more detail?

sunng87:
> Could you describe it in more detail?

Yes, because change like this will update url of something existed articles. In order to prevent 404 for those pages, we will need to automatically redirect user to new url. This behaviour is like github: when you renamed a repo you will be redirect to new url when accessing it from old one.

We will need to be able to configure redirect url when important pages be deleted/renamed.

Aganivi:
get it

zhaoyingnan01 (approved):

sunng87 (approved):

PR #718, opened 2023-12-15
docs: update CONTRIBUTING to use pnpm

tisonkun:
## What's Changed in this PR

We do use pnpm.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,8 +49,8 @@ This project requires Node.js version 18.0.0 or higher.
 Use the following commands to start a local server:
 
 ```shell
-npm install
-npm run dev
+pnpm install
+pnpm run dev
 ```
 
 You can also use `npm run build` to check dead links.


alili:
We all know that pnpm is better, but pnpm is not the default package management tool for Node.js, and it does not rule out that some users are not familiar with the Node.js ecosystem (although this may not be a big deal).

WenyXu:
How about this🥹
```
npm install
npm run dev
# or use pnpm
pnpm install
pnpm run dev
```

tisonkun:
I just notice that we also have a `package-lock.json` file ..

If so, I don't have preference on this doc. But generally we stick on one solution and guide contributors to that way.

> some users are not familiar with the Node.js ecosystem

If we use pnpm primarily, it should be a prerequisite for contributors. NPM runs more slower and when I was a new contributor I don't like it also :D

Closing if we keep each lock file now.

PR #721, opened 2023-12-21
docs: developers generally mean appdevs

tisonkun:
1. Rename "Developer Guide" to "Contributor Guide" because developers generally mean appdevs. They have different meaning for audiences.
2. Move "Contributor Guide" to follow "Reference". We talk everything user aspect and then for contributors.

cc @killme2008 @alili


nicecui:
Thanks and appreciate to @tisonkun for the great idea of this PR. We plan to change the Developer Guide to Contributor Guide in version 0.5 of the docs. I will create a PR which contains this change in the next few days. So maybe we can close this PR 😉 。

tisonkun:
@nicecui Thanks for your feedback! Then I'll understand that you will include the rename **and reorder** in the next version when you cut it.

I'll close this PR as won't do in 0.4 and start a new PR includes other fixups (better links majorly)

PR #738, opened 2023-12-27
fix: range query results

nicecui:
## What's Changed in this PR

update range query results

## Checklist


Taylor-lagrange (commented):
Maybe sorting the outputs by timestamp will make it more readable for users. Although range query output is not guaranteed in order. However, this may cause users to misunderstand that the output of the Range query is ordered.

nicecui:
@Taylor-lagrange All the results have been sorted by `ts` asc. It just doesn’t show `order by ts asc` in the SQL statements.

Wenjie0329 (approved):

Taylor-lagrange (approved):
LGTM

PR #799, opened 2024-01-30
fix: rename files

niebayes:
## What's Changed in this PR

Rename all `kafka_remote_wal.md` to `kafka-remote-wal.md`.

## Checklist


tisonkun (commented):
I can see that we have files with name `order_by.md`, `describe_table.md`, `group_by.md`. Do we try to stick with kebab case for all filename? Or why do you propose this change.

nicecui:
@tisonkun The file names need to be the same as the navigation content in summary.yml (Here is `kafka-remote-wal`), so that the file can be listed in the navigation titles.
We don't need to change `order_by.md`... as it is already the same as the nav in the summary.yml.

nicecui (approved):

Wenjie0329 (approved):

tisonkun:
> The file names need to be the same as the navigation content in summary.yml

Got it. Maybe we can try to stick in one style, said kebab-case.

tisonkun (approved):

PR #801, opened 2024-02-06
docs: ingester-go

yuanbohan:
## What's Changed in this PR

ingester-go documentation

## Checklist


sunng87 (approved):

tisonkun (commented):
Thanks for your updates! May you provide some background on this change? It isn't a new paragraph and I wonder how the previous content becomes wrong or outdated (related PRs or discussions).

yuanbohan:
> Thanks for your updates! May you provide some background on this change? It isn't a new paragraph and I wonder how the previous content becomes wrong or outdated (related PRs or discussions).

Sure.

[greptimedb-client-go](https://github.com/GreptimeTeam/greptimedb-client-go) is the ancestor of [greptimedb-ingester-go](https://github.com/GreptimeTeam/greptimedb-ingester-go), whose interface includes:

- write
- query via sql and promql
- metric & series model

but greptimedb sdk won't support query any more, and metric & series model should be replaced by Table & Column & Row. So what greptimedb-ingester-go includes:

- interface not compatible with greptimedb-client-go
- similar interface with [greptimedb-ingester-java](https://github.com/GreptimeTeam/greptimedb-ingester-java): Table, Column and Row model.
- Create struct or slice of structs into GreptimeDB directly without Table model, similar with [WritePoJo in greptimedb-ingester-java](https://github.com/GreptimeTeam/greptimedb-ingester-java/blob/main/ingester-example/src/main/java/io/greptime/WritePOJOsQuickStart.java)

Anyc66666666 (approved):

tisonkun (approved):
Thanks for your information! LGTM.

I wonder if other clients would become ingester also - this may change our concepts exposed.

yuanbohan:
#803 has covered the details

Issue #832, opened 2024-03-11
Document how RANGE QUERY supports WHERE


tisonkun:
### What problem does the new feature solve?

```sql
SELECT ts, COUNT(1) RANGE '1h' as times FROM keyboard_monitor ALIGN '1h' WHERE date_trunc('day', ts) = '2024-03-11' ORDER BY ts DESC LIMIT 10;
```

It throws:

```
ERROR 1815 (HY000): SQL statement is not supported: SELECT ts, COUNT(1) RANGE '1h' as times FROM keyboard_monitor ALIGN '1h' WHERE date_trunc('day', ts) = '2024-03-11' ORDER BY ts DESC LIMIT 10, keyword: WHERE
```

Although I can work it around by:

```sql
SELECT 
    ts,
    COUNT(1) RANGE '1h' as times
FROM 
    (
        SELECT ts
        FROM keyboard_monitor
        WHERE date_trunc('day', ts, 'Asia/Shanghai') = '2024-03-11'
    )
    ALIGN '1h'
ORDER BY ts DESC LIMIT 10;
```

I wonder if RANGE QUERY can support WHERE filters; or it's theoretically impossible?

### What does the feature do?

Support `WHERE` in RANGE QUERY.

PR #850, opened 2024-03-20
docs: update range query docs

Taylor-lagrange:
## What's Changed in this PR

update range query docs
- new fill behavior
- Range SQL structure
- `first_value`/`last_value` use `ts` as default `order by` 

related pr: https://github.com/GreptimeTeam/greptimedb/pull/3489

## Checklist

Closes https://github.com/GreptimeTeam/docs/issues/832

### DIFF
--- a/docs/v0.7/en/reference/sql/range.md
+++ b/docs/v0.7/en/reference/sql/range.md
@@ -26,7 +26,7 @@ INTERVAL :=  TIME_INTERVAL | ( INTERVAL expr )
    - Strings based on the `PromQL Time Durations` format (eg: `3h`, `1h30m`). Visit the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations) for a more detailed description of this format.
    - `Interval` type. To use the `Interval` type, you need to carry parentheses, (for example: `(INTERVAL '1 year 3 hours 20 minutes')`). Visit [Interval](./functions.md#interval) for a more detailed description of this format.
 - `AGGR_FUNCTION(column1, column2,..) RANGE INTERVAL [FILL FILL_OPTION]` is called a Range expression.
-   - `AGGR_FUNCTION(column1, column2,..)` is an aggregate function that represents the expression that needs to be aggregated. Aggregate functions support the use of `order by` expressions. (Note: `first_value` and `last_value` functions use timestamp as `order by` expression by default)
+   - `AGGR_FUNCTION(column1, column2,..)` is an aggregate function that represents the expression that needs to be aggregated.
    - Keyword `RANGE`, required field, followed by parameter `INTERVAL` specifies the time range of each data aggregation.
    - Keyword `FILL`, optional field, please see [`FILL` Option](#fill-option) for details.
    - Range expressions can be combined with other operations to implement more complex queries. For details, see [Nested Range Expressions](#nested-range-expressions).
@@ -367,6 +367,56 @@ Get after running
 | 2023-01-02 00:59:55 |                                 1 |
 | 2023-01-02 01:00:00 |                                 1 |
 +---------------------+-----------------------------------+
+```
+
+## `ORDER BY` option
+
+Range queries support the use of `order by` expressions within specific aggregate functions to sort data on the same time slot. Taking the `first_value` and `last_value` aggregate functions as an example, users can use `order by` to sort the aggregated data to correctly select the first value and the last value. By default, the `first_value` and `last_value` functions sort data in ascending order using timestamps.
+
+Take this table as an example:
+
+```sql
++---------------------+-------+------+-------+
+| ts                  | host  | val  | addon |
++---------------------+-------+------+-------+
+| 1970-01-01 00:00:00 | host1 |    0 |     3 |
+| 1970-01-01 00:00:01 | host1 |    1 |     2 |
+| 1970-01-01 00:00:02 | host1 |    2 |     1 |
++---------------------+-------+------+-------+
+```
+
+Users can not specify the `order by` expression, and the effect is equivalent to using the `ts` column to sort in ascending order.
+
+```sql
+SELECT ts, first_value(val) RANGE '5s', last_value(val) RANGE '5s' FROM host ALIGN '5s';
+-- Equivalent to
+SELECT ts, first_value(val order by ts ASC) RANGE '5s', last_value(val order by ts ASC) RANGE '5s' FROM host ALIGN '5s';
+```
+
+Get after query
+
+```sql
++---------------------+--------------------------------+-------------------------------+
+| ts                  | FIRST_VALUE(host.val) RANGE 5s | LAST_VALUE(host.val) RANGE 5s |
++---------------------+--------------------------------+-------------------------------+
+| 1970-01-01 00:00:00 |                              0 |                             2 |
++---------------------+--------------------------------+-------------------------------+
+```
+
+Users can also specify their own sorting rules without using the default sorting options, such as using `addon` sorting:
+
+```sql
+SELECT ts, first_value(val ORDER BY addon ASC) RANGE '5s', last_value(val ORDER BY addon ASC) RANGE '5s' FROM host ALIGN '5s';
+```
+
+Get after query
+
+```sql
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
+| ts                  | FIRST_VALUE(host.val) ORDER BY [host.addon ASC NULLS LAST] RANGE 5s | LAST_VALUE(host.val) ORDER BY [host.addon ASC NULLS LAST] RANGE 5s |
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
+| 1970-01-01 00:00:00 |                                                                   2 |                                                                  0 |
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
 ```
 
 ## Nested Range Expressions
--- a/docs/v0.7/en/reference/sql/range.md
+++ b/docs/v0.7/en/reference/sql/range.md
@@ -41,15 +41,15 @@ The `FILL` keyword can also be used after the `ALIGN` keyword to specify the def
 if no fill option is provided.
 
 For example, in the following SQL code,
-the `max(cpu) RANGE '10s'` Range expression uses the fill option `LINEAR`, while the `min(cpu) RANGE '10s'` Range expression,
+the `max(val) RANGE '10s'` Range expression uses the fill option `LINEAR`, while the `min(val) RANGE '10s'` Range expression,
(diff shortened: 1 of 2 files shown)


docs/v0.7/en/reference/sql/range.md
```diff
@@ -23,15 +26,15 @@ INTERVAL :=  TIME_INTERVAL | ( INTERVAL expr )
    - Strings based on the `PromQL Time Durations` format (eg: `3h`, `1h30m`). Visit the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations) for a more detailed description of this format.
    - `Interval` type. To use the `Interval` type, you need to carry parentheses, (for example: `(INTERVAL '1 year 3 hours 20 minutes')`). Visit [Interval](./functions.md#interval) for a more detailed description of this format.
 - `AGGR_FUNCTION(column1, column2,..) RANGE INTERVAL [FILL FILL_OPTION]` is called a Range expression.
-   - `AGGR_FUNCTION(column1, column2,..)` is an aggregate function that represents the expression that needs to be aggregated.
+   - `AGGR_FUNCTION(column1, column2,..)` is an aggregate function that represents the expression that needs to be aggregated. Aggregate functions support the use of `order by` expressions. (Note: `first_value` and `last_value` functions use timestamp as `order by` expression by default)
```
nicecui, line 29: I don't understand what NOTE means. Does that mean the type of the column specified by `order by` must be a timestamp? Does that mean that other aggregate functions can use any type of column as an order by parameter?

docs/v0.7/en/reference/sql/range.md
```diff
@@ -60,74 +63,135 @@ ALIGN '5s' BY (host) FILL PREV;
 | `LINEAR` | Use [linear interpolation](https://en.wikipedia.org/wiki/Linear_interpolation) to fill the data. If an integer type is filled with `LINEAR`, the variable type of the column will be implicitly converted to a floating point type during calculation |
 |   `X`    |                                           Fill in a constant, the data type of the constant must be consistent with the variable type of the Range expression                                            |
 
+:::tip NOTE
```
nicecui, line 66: A note shouldn't be too long, and please avoid using too many notes.
Here seems to be a special case of the `FILL` option, so please remove the `:::tip NOTE` and put it at the bottom of the introduction of the `FILL` option.

docs/v0.7/en/reference/sql/range.md
```diff
@@ -60,74 +63,135 @@ ALIGN '5s' BY (host) FILL PREV;
...
+ALIGN '5s';
+```
+:::
 
 Take the following table as an example:
 
 ```sql
 +---------------------+-------+------+
-| ts                  | host  | cpu  |
+| ts                  | host  | val  |
```
nicecui, line 96: If you are using `val` as the column name, please rename all `cpu` column to `val` column in the section of the `FILL` option, including SQLs and results.

docs/v0.7/en/reference/sql/range.md
```diff
@@ -60,74 +63,135 @@ ALIGN '5s' BY (host) FILL PREV;
...
 +---------------------+-------+------+
-| 2023-01-01 08:00:00 | host1 |  4.5 |
-| 2023-01-01 08:00:05 | host1 | NULL |
-| 2023-01-01 08:00:10 | host1 |  6.5 |
+| 1970-01-01 00:00:00 | host1 |    0 |
+| 1970-01-01 00:00:15 | host1 |    6 |
+| 1970-01-01 00:00:00 | host2 |    6 |
+| 1970-01-01 00:00:15 | host2 |   12 |
```
nicecui, line 101: The example should have a row with value `NULL` to show the behavior of the different `FILL` options.

docs/v0.7/en/reference/sql/range.md
```diff
@@ -60,74 +63,135 @@ ALIGN '5s' BY (host) FILL PREV;
 | `LINEAR` | Use [linear interpolation](https://en.wikipedia.org/wiki/Linear_interpolation) to fill the data. If an integer type is filled with `LINEAR`, the variable type of the column will be implicitly converted to a floating point type during calculation |
 |   `X`    |                                           Fill in a constant, the data type of the constant must be consistent with the variable type of the Range expression                                            |
 
+:::tip NOTE
+Note that if there are multiple Range expressions and the FILL method is specified for only one of them, in order to keep the number of SQL output rows uniform, the other Range expressions will use the FILL NULL method to fill in the missing time slots. So the following two SQL statements are equivalent in output:
```
nicecui, line 67:
```suggestion
If there are multiple Range expressions but only one of them specifies the `Fill` option, the other Range expressions will use the `FILL NULL` method to fill in the missing time slots.
The following two SQL statements are equivalent:
```

nicecui (commented):
Please also modify the zh document according to the comments, thank you very much!

docs/v0.7/en/reference/sql/range.md
Taylor-lagrange: The `first_value` aggregate function is used to find the first value in an aggregation group. It generally needs to be used with `order by` to specify what the first value is. In range queries, timestamps are widely used as the sorting basis. Therefore, in order to simplify the writing of SQL, the default `order by` parameter of `first_value` is timestamp, and the same for `last_value`.

docs/v0.7/en/reference/sql/range.md
Taylor-lagrange: It's the new behavior of FILL option, see pr https://github.com/GreptimeTeam/greptimedb/pull/3489

docs/v0.7/en/reference/sql/range.md
```diff
@@ -308,6 +369,56 @@ Get after running
 +---------------------+-----------------------------------+
 ```
 
+## `ORDER BY` option
+
+Range queries support the use of `order by` expressions within specific aggregate functions to sort data on the same time slot. Taking the `first_value` and `last_value` aggregate functions as an example, users can use `order by` to sort the aggregated data to correctly select the first value and the last value. By default, the `first_value` and `last_value` functions sort data in ascending order using timestamps.
```
nicecui, line 374:
```suggestion
Range queries support the use of `order by` expressions in the parameters of the `first_value` and `last_value` aggregate functions. By default, the data is sorted in ascending order based on the time index column.
```

docs/v0.7/en/reference/sql/range.md
```diff
@@ -308,6 +369,56 @@ Get after running
...
++---------------------+-------+------+-------+
+| ts                  | host  | val  | addon |
++---------------------+-------+------+-------+
+| 1970-01-01 00:00:00 | host1 |    0 |     3 |
+| 1970-01-01 00:00:01 | host1 |    1 |     2 |
+| 1970-01-01 00:00:02 | host1 |    2 |     1 |
++---------------------+-------+------+-------+
+```
+
+Users can not specify the `order by` expression, and the effect is equivalent to using the `ts` column to sort in ascending order.
```
nicecui, line 388:
```suggestion
When the `order by` expression is not specified in the function parameter, the default behavior is to sort the data in ascending order based on the time index column.
```

docs/v0.7/en/reference/sql/range.md
```diff
@@ -308,6 +369,56 @@ Get after running
...
+
+```sql
++---------------------+--------------------------------+-------------------------------+
+| ts                  | FIRST_VALUE(host.val) RANGE 5s | LAST_VALUE(host.val) RANGE 5s |
++---------------------+--------------------------------+-------------------------------+
+| 1970-01-01 00:00:00 |                              0 |                             2 |
++---------------------+--------------------------------+-------------------------------+
+```
+
+Users can also specify their own sorting rules without using the default sorting options, such as using `addon` sorting:
```
nicecui, line 406:
```suggestion
You can specify your own sorting rules. For example, the following SQL sorts the data by the `addon` column in ascending order:
```

docs/v0.7/en/reference/sql/range.md
```diff
@@ -308,6 +369,56 @@ Get after running
...
+Get after query
+
+```sql
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
+| ts                  | FIRST_VALUE(host.val) ORDER BY [host.addon ASC NULLS LAST] RANGE 5s | LAST_VALUE(host.val) ORDER BY [host.addon ASC NULLS LAST] RANGE 5s |
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
+| 1970-01-01 00:00:00 |                                                                   2 |                                                                  0 |
++---------------------+---------------------------------------------------------------------+--------------------------------------------------------------------+
+```
+
```
nicecui, line 421: What is the behavior of the following SQL:
```
SELECT ts, first_value(val ORDER BY addon ASC) RANGE '5s', last_value(val ORDER BY addon ASC) RANGE '5s' FROM host ALIGN '5s' ORDER BY addon DESC;
```
ORDER BY also added at the end of the SQL and the specified behavior is different from the former.

nicecui (commented):
Please also modify the zh document according to the comments, thank you very much

docs/v0.7/en/reference/sql/range.md
nicecui: What is the behavior of the SQL that only specified `order by` at the end of the SQL
```
SELECT ts, first_value(val) RANGE '5s', last_value(val) RANGE '5s' FROM host ALIGN '5s' ORDER BY addon DESC;
```

Taylor-lagrange: - `first_value(val ORDER BY addon ASC) RANGE '5s'` means sort data in aggragate group, so we can choose our desire value in `first_value`.
- order by in SQL level meas sort the output of entire output data.

docs/v0.7/zh/reference/sql/range.md
```diff
@@ -304,6 +366,57 @@ FROM host ALIGN '5s' BY ();
 +---------------------+-----------------------------------+
 ```
 
+## 聚合函数中的 `ORDER BY` 选项
+
+Range 查询支持在聚合函数 `first_value` 和 `last_value` 中使用 `order by` 表达式，默认情况下，聚合函数使用时间列升序排列数据。
```
nicecui, line 371:
```suggestion
Range 查询支持在聚合函数 `first_value` 和 `last_value` 中使用 `order by` 表达式，默认情况下，聚合函数使用时间索引列升序排列数据。
```

nicecui (approved):

tisonkun (approved):

tisonkun (commented):
@Taylor-lagrange @nicecui we should update the `FILL` table to avoid

```
|  `NULL`  |                                                                                Fill directly with `NULL` (default method)                                                                                |
```

Since the default is now "NO FILL"

PR #854, opened 2024-03-25
docs: timezone setting

Taylor-lagrange:
## What's Changed in this PR

Docs about how to use set timezone under different protocol, and what influence time zone will bring in query and insert

https://github.com/GreptimeTeam/greptimedb/issues/2907

## Checklist


nicecui (approved):

docs/v0.7/en/user-guide/clients/http-api.md
```diff
@@ -49,11 +49,55 @@ http://localhost:4000/v1/sql?db=public
 InfluxDB uses its own authentication format, see [InfluxDB](./influxdb-line.md) for details.
 :::
 
-## Write Data
+## Time zone
+
+GreptimeDB supports the `X-Greptime-Timezone` field in the header of the HTTP protocol.
```
tisonkun, line 54:
```suggestion
GreptimeDB supports the `X-Greptime-Timezone` header in HTTP requests.
```

IIRC HTTP Xxx header is a formal description. cc @nicecui  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type "a Content-Type header"

docs/v0.7/en/user-guide/clients/http-api.md
```diff
@@ -49,11 +49,55 @@ http://localhost:4000/v1/sql?db=public
...
+It is used to specify the timezone for the current SQL query.
+
+For example, the following request uses the time zone `+1:00` for the query:
```
tisonkun, line 57:
```suggestion
For example, the following request uses the timezone `+1:00` for the query:
```

If we use `timezone` consistently, let's keep it.

killme2008:
@tisonkun PTAL

tisonkun:
Seems "time zone" the proper expression. I'll update.

tisonkun (approved):

PR #858, opened 2024-03-26
docs: copy database example

### LINKED ISSUE #857: Update docs for feat: support time range in copy table
A document change request is generated from
https://github.com/GreptimeTeam/greptimedb/pull/3583

v0y4g3r:
## What's Changed in this PR

Add time range option doc for `COPY` statements.

Closes #857
## Checklist

- [X] Please ensure that the content in `summary.yml` matches the current document structure when you changed the document structure.

### DIFF
--- a/docs/v0.6/en/reference/sql/copy.md
+++ b/docs/v0.6/en/reference/sql/copy.md
@@ -23,7 +23,7 @@ data (`/xxx/xxx/output.parquet` in this case).
 | Option  | Description  | Required |
 |---|---|---|
 | `FORMAT` | Target file(s) format, e.g., JSON, CSV, Parquet  | **Required** |
-| `START_TIME`/`END_TIME`| The time range within which data should be exported | Optional |
+| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive. | Optional |
 
 #### `CONNECTION` Option
 
--- a/docs/v0.6/en/reference/sql/copy.md
+++ b/docs/v0.6/en/reference/sql/copy.md
@@ -142,7 +142,7 @@ COPY DATABASE <db_name>
 | Option  | Description  | Required |
 |---|---|---|
 | `FORMAT` | Export file format, available options: JSON, CSV, Parquet  | **Required** |
-| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive) | Optional |
+| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive. | Optional |
 
 > - When copying databses, `<PATH>` must end with `/`.
 > - `CONNECTION` parameters can also be used to copying databases to/from object storage services like AWS S3.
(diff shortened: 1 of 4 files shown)


docs/v0.6/en/reference/sql/copy.md
```diff
@@ -135,9 +139,26 @@ COPY DATABASE <db_name>
   )]
 ```
 
-- When copying databases, `<PATH>` must end with `/`.
-- `FORMAT` option has the same available values as copying tables.
-- `CONNECTION` parameters can also be used to copying databases to/from object storage services like [AWS S3](#connect-to-s3).
+| Option  | Description  | Required |
+|---|---|---|
+| `FORMAT` | Export file format, available options: JSON, CSV, Parquet  | **Required** |
+| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive) | Optional |
```
nicecui, line 145:
```suggestion
| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive. | Optional |
```

docs/v0.6/en/reference/sql/copy.md
```diff
@@ -23,6 +23,7 @@ data (`/xxx/xxx/output.parquet` in this case).
 | Option  | Description  | Required |
 |---|---|---|
 | `FORMAT` | Target file(s) format, e.g., JSON, CSV, Parquet  | **Required** |
+| `START_TIME`/`END_TIME`| The time range within which data should be exported | Optional |
```
nicecui, line 26:
```suggestion
| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive. | Optional |
```

nicecui (commented):
Can you move the content to `docs/v0.7`? Thank you very much!

v0y4g3r:
> Can you move the content to `docs/v0.7`? Thank you very much!

I copied the docs related to `copy` under v0.6  to v0.7 directory since it's applicable to both version.

nicecui (approved):

docs/v0.6/en/reference/sql/copy.md
```diff
@@ -23,6 +23,7 @@ data (`/xxx/xxx/output.parquet` in this case).
 | Option  | Description  | Required |
 |---|---|---|
 | `FORMAT` | Target file(s) format, e.g., JSON, CSV, Parquet  | **Required** |
+| `START_TIME`/`END_TIME`| The time range within which data should be exported. `START_TIME` is inclusive and `END_TIME` is exclusive. | Optional |
```
tisonkun, line 26: What happen if START_TIME <= END_TIME?

I suppose it would be an empty selection, but will it generate a warning?

v0y4g3r: > What happen if START_TIME <= END_TIME?
> 
> I suppose it would be an empty selection, but will it generate a warning?

Currently it will create an invalid timestamp range, thus it will not come into effect. But I think maybe we should reject this statement instead?

tisonkun: Silently invalid value is error-prone.

At least a warning and clearly fallback to empty range (empty to none). But better to be an error.

Not a blocker for merging this patch. Let's track an issue at the DB repo.

tisonkun (approved):

docs/v0.6/en/reference/sql/copy.md
v0y4g3r: > Silently invalid value is error-prone.
> 
> At least a warning and clearly fallback to empty range (empty to none). But better to be an error.
> 
> Not a blocker for merging this patch. Let's track an issue at the DB repo.

I will draft a pr to address this issue.

PR #955, opened 2024-05-15
build: use baseurl in version switcher

tisonkun:
Try to preview all the versions more smoothly on Cloudflare pages.


tisonkun:
Different versions stay in different SPA and thus they have different baseUrl. It's hard to do in page nav but we should find a way to pass the host and construct the external link with the deploy host url.

PR #1107, opened 2024-07-29
docs: remove docs for our promql path

sunng87:
## What's Changed in this PR

Removes our own `/promql` API from docs to avoid misunderstanding.

## Checklist


killme2008:
I think it's not confusing currently because the documents describe it clearly. We'll refactor the user guide, leaving it right now would be good.

docs/nightly/en/user-guide/query-data/promql.md
```diff
@@ -24,98 +24,8 @@ Consult [Prometheus
 documents](https://prometheus.io/docs/prometheus/latest/querying/api) for usage
 of these API.
 
-You can use additional query parameter `db` to specify GreptimeDB database name.
-
-## GreptimeDB's HTTP API
```
killme2008, line 29: Don't delete them,just comment them.

PR #1137, opened 2024-08-19
docs: add performance tuning tips

evenyag:
## What's Changed in this PR

This document contains some tips for tuning GreptimeDB performance.

## Checklist


docs/user-guide/operations/performance-tuning-tips.md
```diff
@@ -0,0 +1,161 @@
...
+|---|---|---|
+| greptime_mito_read_stage_elapsed_bucket | histogram | The elapsed time of different phases of a query in the storage engine. |
+| greptime_mito_cache_bytes | gauge | Size of cached contents |
+| greptime_mito_cache_hit | counter | Total count of cache hit |
+| greptime_mito_cache_miss | counter | Total count of cache miss |
+
+
+### Using cache for object stores
+
+While using object stores, it's highly recommended to enable the object store read cache and the write cache in the storage engine. This could reduce query time by more than 10 times.
```
nicecui, line 22:
```suggestion
It's highly recommended to enable the object store read cache and the write cache in the storage engine. This could reduce query time by more than 10 times.
```

Users use object stores, there is no `while`.

docs/user-guide/operations/performance-tuning-tips.md
```diff
@@ -0,0 +1,161 @@
...
+# experimental_write_cache_path = "/path/to/write/cache"
+```
+
+### Enlarging cache size
+
+The `greptime_mito_cache_bytes` and `greptime_mito_cache_miss` metrics can indicate whether we need to enlarge the cache size:
+- The `greptime_mito_cache_miss` is high and increasing
+- The `greptime_mito_cache_bytes` reaches the cache capacity
+
+The `type` label in these metrics shows the type of cache. The storage engine supports configuring the size of each cache. The following is an example:
```
nicecui, line 63:
```suggestion
You can monitor the `greptime_mito_cache_bytes` and `greptime_mito_cache_miss` metrics to determine if you need to increase the cache size. The `type` label in these metrics indicates the type of cache.

If the `greptime_mito_cache_miss` metric is consistently high and increasing, or if the `greptime_mito_cache_bytes` metric reaches the cache capacity, you may need to adjust the cache size configurations of the storage engine.

Here's an example:
```

docs/user-guide/operations/performance-tuning-tips.md
```diff
@@ -0,0 +1,161 @@
...
+
+```toml
+[[region_engine]]
+[region_engine.mito]
+scan_parallelism = 8
+```
+
+### Using append-only table if possible
+
+In general, append-only tables have a higher scan performance as the storage engine can skip merging and deduplication. We recommend enabling the [append_mode](/reference/sql/create.md#create-an-append-only-table) for the table if it doesn't require deduplication. For example, a log table should be append-only as log messages may have the same timestamp.
```
nicecui, line 87: hmm... append-only is not for performance optimization, but for the needs of specific business scenarios.
So it might not be appropriate to put it here.

nicecui (commented):
LGTM

docs/user-guide/operations/performance-tuning-tips.md
evenyag: It is not only for specific business scenarios. It may improve the performance as it skips deduplication so it can scan faster. We also have some optimize rules for append-only tables. (For example, `count(*)` doesn't require reading all rows`).

I'll add these to the doc.

docs/user-guide/operations/performance-tuning-tips.md
```diff
@@ -0,0 +1,163 @@
...
+Here's an example:
+
+```toml
+[[region_engine]]
+[region_engine.mito]
+experimental_write_cache_size = "10G"
+sst_meta_cache_size = "128MB"
+vector_cache_size = "512MB"
+page_cache_size = "512MB"
+selector_result_cache_size = "512MB"
```
killme2008, line 72: Please explain all these cache settings. What do they mean, and what are the recommended values?

killme2008 (commented):
Good job

docs/user-guide/operations/performance-tuning-tips.md
```diff
@@ -0,0 +1,163 @@
...
+|---|---|---|
+| greptime_mito_read_stage_elapsed_bucket | histogram | The elapsed time of different phases of a query in the storage engine. |
+| greptime_mito_cache_bytes | gauge | Size of cached contents |
+| greptime_mito_cache_hit | counter | Total count of cache hit |
+| greptime_mito_cache_miss | counter | Total count of cache miss |
+
+
+### Using cache for object stores
+
+It's highly recommended to enable the object store read cache and the write cache in the storage engine. This could reduce query time by more than 10 times.
```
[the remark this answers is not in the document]
evenyag: The write buffer usually means the memtable and this is a write-through cache.

nicecui (commented):
please also add the zh document to `/Users/yiran/dev/docs/i18n/zh/docusaurus-plugin-content-docs/current` after all the English comments resolved, thank you

docs/user-guide/operations/performance-tuning-tips.md
evenyag: done

evenyag:
Still can't find the page newly added in the preview URL
https://9281b489.greptime-docs.pages.dev/

nicecui (approved):

PR #1144, opened 2024-08-21
docs: update the CONTRIBUTING.md

nicecui:
## What's Changed in this PR

as title

## Checklist


src/misc/update_functions.rb
```diff
@@ -0,0 +1,78 @@
...
+      line = process_headlines(line)
+      line = fix_links(line)
+      line
+    end
+
+    f.puts lines.join("\n")
+  end
+end
+
+File.rename "temp.md", "./docs/reference/sql/functions/df-functions.md"
```
nicecui, line 76: @killme2008 PTAL,
I removed the parameters here because it is too complicated to maintain old versions of the documentation on Docusaurus.

sunchanglong (approved):

sunng87 (approved):

PR #1176, opened 2024-09-06
chore: remove dissect dynamic key names

paomian:
## What's Changed in this PR

> remove dissect reference keys. because the resulting key is uncertain, it can't be specified by transform; if it is certain, you can use name matching to get the specified key directly.

* remove dissect dynamic key names. referable https://github.com/GreptimeTeam/greptimedb/pull/4571

## Checklist


nicecui (commented):
LGTM, please also update the locale doc at `/i18n/zh/docusaurus-plugin-content-docs/current`

paomian:
> LGTM, please also update the locale doc in `/i18n/zh/docusaurus-plugin-content-docs/current`

done.

nicecui (approved):

shuiyisong (approved):

Issue #1145, opened 2024-08-21
Fulltext case-sensitive index behavior


gar1t:
In [Quick Start](https://docs.greptime.com/getting-started/quick-start/), the query:

```sql
SELECT 
  ts,
  api_path,
  log
FROM
  app_logs
WHERE
  matches(log, 'timeout');
```

shows results that are case-sensitive:

```
+---------------------+------------------+--------------------+
| ts                  | api_path         | log                |
+---------------------+------------------+--------------------+
| 2024-07-11 20:00:10 | /api/v1/billings | Connection timeout |
| 2024-07-11 20:00:10 | /api/v1/resource | Connection timeout |
+---------------------+------------------+--------------------+
2 rows in set (0.01 sec)
```

However, the table def is this:

```
Create Table: CREATE TABLE IF NOT EXISTS `app_logs` (
...
`log` STRING NULL FULLTEXT WITH(analyzer = 'English', case_sensitive = 'false'),
...)
```

The docs for [CREATE](https://docs.greptime.com/reference/sql/create/) indicate that `case_sensitive` for `FULLTEXT` is `true`. Based on what I'm seeing, following Quick Start, the default is `false`.

In any event, the query behavior is case sensitive.

Issues as I see them:

- Possible error in either docs or implementation for default value of `case_sensitive` for fulltext index
- Case-sensitive match behavior when schema shows `case_sensitive` to be `false`

zhongzc:
Thank you for your thorough review; the issue does indeed exist.

The specific reason is that the calculation for `matches` is separate between frontend and datanode. Datanode does respect the case-sensitive configuration, but this part has not yet been completed in frontend (see TODO): https://github.com/GreptimeTeam/greptimedb/blob/9c1704d4cbbfab8af07a77da598a1cfe2a5e7b22/src/common/function/src/scalars/matches.rs#L75-L95. As it stands, the implementation is currently case-sensitive.

Therefore, until this part of the work is completed, to maintain consistency, I think we can either hardcode this configuration to true and make it unchangeable, or hardcode it to false, but then change https://github.com/GreptimeTeam/greptimedb/blob/9c1704d4cbbfab8af07a77da598a1cfe2a5e7b22/src/common/function/src/scalars/matches.rs#L205 to use `ilike`, which would be more practical.

In any case, it was indeed an oversight, and I will arrange for a prompt fix.

cc @waynexia

PR #1271, opened 2024-11-07
docs: update alter column fulltext options

### LINKED ISSUE #1277: Add docs for altering fulltext options
https://github.com/GreptimeTeam/greptimedb/pull/4952

CookiePieWw:
## What's Changed in this PR

Add docs for alter column fulltext options. Waiting for https://github.com/GreptimeTeam/greptimedb/pull/4952

Close #1277 

## Checklist

### DIFF
--- a/docs/reference/sql/alter.md
+++ b/docs/reference/sql/alter.md
@@ -12,6 +12,7 @@ ALTER TABLE [db.]table
    [ADD COLUMN name type [options]
     | DROP COLUMN name
     | MODIFY COLUMN name type
+    | MODIFY COLUMN name SET FULLTEXT [WITH <options>]
     | RENAME name
     | SET <option_name>=<option_value> [, ...]
    ]
@@ -76,6 +77,27 @@ Currently following options are supported:
 ```sql
 ALTER TABLE monitor SET 'ttl'='1d';
 ```
+
+### Modify column fulltext index options
+
+Modify the fulltext index options of a column
+
+```sql
+ALTER TABLE monitor MODIFY COLUMN load_15 SET FULLTEXT WITH (enable = 'true', analyzer = 'Chinese', case_sensitive = 'false');
+```
+
+You can specify the following options using `FULLTEXT WITH`:
+
+- `enable`: Determines whether the full-text index is enabled. Supported values are `true` and `false`. Default is `true`.
+- `analyzer`: Sets the language analyzer for the full-text index. Supported values are `English` and `Chinese`. Default is `English`.
+- `case_sensitive`: Determines whether the full-text index is case-sensitive. Supported values are `true` and `false`. Default is `false`.
+
+If `WITH` is not specified, `FULLTEXT` will use the default values.
+
+Currently, the fulltext index can be altered when:
+
+1. The column is of a string type.
+2. Either the column has no fulltext index and the alter command is to enable it, or the column has a fulltext index and the alter command is to disable it.
 
 ### Rename table
 
--- a/docs/reference/sql/alter.md
+++ b/docs/reference/sql/alter.md
@@ -13,6 +13,7 @@ ALTER TABLE [db.]table
     | DROP COLUMN name
     | MODIFY COLUMN name type
     | MODIFY COLUMN name SET FULLTEXT [WITH <options>]
+    | MODIFY COLUMN name UNSET FULLTEXT
     | RENAME name
     | SET <option_name>=<option_value> [, ...]
    ]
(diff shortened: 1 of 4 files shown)


docs/reference/sql/create.md
```diff
@@ -268,7 +268,7 @@ You can specify the following options using `FULLTEXT WITH`:
 If `WITH` is not specified, `FULLTEXT` will use the following default values:
 
 - `analyzer`: default is `English`
-- `case_sensitive`: default is `true`
+- `case_sensitive`: default is `false`
```
CookiePieWw, line 271: Not sure if it's alright to change it because of #1145. cc @zhongzc.

docs/reference/sql/alter.md
```diff
@@ -77,6 +78,27 @@ Currently following options are supported:
 ALTER TABLE monitor SET 'ttl'='1d';
 ```
 
+### Modify column fulltext index options
+
+Modify the fulltext index options of a column
```
nicecui, line 83:
```suggestion
Modify the fulltext index options of a column.
```

i18n/zh/docusaurus-plugin-content-docs/current/reference/sql/alter.md
```diff
@@ -77,6 +78,27 @@ ALTER TABLE monitor MODIFY COLUMN load_15 STRING;
 ALTER TABLE monitor SET 'ttl'='1d';
 ```
 
+### 修改列全文索引选项
+
+修改列的全文索引选项
```
nicecui, line 83:
```suggestion
修改列的全文索引选项。
```

docs/reference/sql/alter.md
```diff
@@ -77,6 +78,27 @@ Currently following options are supported:
...
+- `enable`: Determines whether the full-text index is enabled. Supported values are `true` and `false`. Default is `true`.
+- `analyzer`: Sets the language analyzer for the full-text index. Supported values are `English` and `Chinese`. Default is `English`.
+- `case_sensitive`: Determines whether the full-text index is case-sensitive. Supported values are `true` and `false`. Default is `false`.
+
+If `WITH` is not specified, `FULLTEXT` will use the default values.
+
+Currently, the fulltext index can be altered when:
+
+1. The column is of a string type.
+2. Either the column has no fulltext index and the alter command is to enable it, or the column has a fulltext index and the alter command is to disable it.
```
nicecui, line 100: It seems the meaning here is different from that in the zh doc

nicecui (approved):

docs/reference/sql/alter.md
```diff
@@ -77,6 +78,26 @@ Currently following options are supported:
...
+ALTER TABLE monitor MODIFY COLUMN load_15 SET FULLTEXT WITH (enable = 'true', analyzer = 'Chinese', case_sensitive = 'false');
+```
+
+You can specify the following options using `FULLTEXT WITH`:
+
+- `enable`: Determines whether the full-text index is enabled. Supported values are `true` and `false`. Default is `true`.
+- `analyzer`: Sets the language analyzer for the full-text index. Supported values are `English` and `Chinese`. Default is `English`.
+- `case_sensitive`: Determines whether the full-text index is case-sensitive. Supported values are `true` and `false`. Default is `false`.
+
+If `WITH` is not specified, `FULLTEXT` will use the default values.
```
zhongzc, line 95:
```suggestion
If `WITH <options>` is not specified, `FULLTEXT` will use the default values.
```

zhongzc (approved):
rest LGTM

docs/reference/sql/alter.md
```diff
@@ -77,6 +79,31 @@ Currently following options are supported:
...
+
+Disable fulltext index on a column:
+
+```sql
+ALTER TABLE monitor MODIFY COLUMN load_15 UNSET FULLTEXT;
+```
+
+The column must be a string type to alter the fulltext index.
+
+When the fulltext index is disabled, you can enable it and specify the `analyzer` and `case_sensitive` options. When the fulltext index is already enabled on a column, you can disable it but cannot modify the options.
```
WenyXu, line 105: What will happen if users disable the full-text index and enable it with different options? @zhongzc

CookiePieWw: For `UNSET` part, the previously constructed index would still take effect. For `SET` part, I guess the engine will start to construct the index and apply it once finished. cc @zhongzc for validation.

zhongzc: Oh, in principle, it should not be allowed to alter options other than `enable`.

After modification, new data will use different options from old data to build indexes, which may produce some unexpected query results.

If the current implementation allows altering options other than "enable", this can be pointed out in the documentation.

PR #1388, opened 2024-12-18
docs: update dryrun docs

paomian:
## What's Changed in this PR

* update dryrun doc

## Checklist


sunng87 (commented):
I'm afraid this example is more complex than the api itself. we can just demo the input data structure here and say this api is designed for dashboard usage.

nicecui:
outdated

PR #1590, opened 2025-03-20
docs: rewrite the data model and schema design docs

evenyag:
## What's Changed in this PR

This PR rewrites the data model and schema design docs.

The new design guide covers
- Column types and selection
- Primary key and index selection
- Deduplication
- Distributed tables

It also modifies the data model document to correct some concepts.

## Checklist


docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
 
-Tag columns, also known as label columns,
-generally carry metadata of the measured data or logs/events.
-For example, when collecting nationwide weather temperature data,
-the city (e.g., `city="New York"`) is a typical tag column.
-In monitoring, system metrics like CPU and memory usually involve a `host` tag to represent the hostname.
+1. Defining the ordering of data in storage.
+   GreptimeDB reuses the `PRIMARY KEY` constraint to define tag columns and the ordering of tags.
+   Unlike traditional databases, in GreptimeDB, primary key and time series are identical.
+   Tables in GreptimeDB sort rows by (primary key, timestamp).
```
killme2008, line 41:
```suggestion
   Tables in GreptimeDB sort rows by in the order of `(primary key, timestamp)`.
```

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
+In GreptimeDB, tag columns are optional. The main purposes of Tag columns include:
 
-Tag columns, also known as label columns,
-generally carry metadata of the measured data or logs/events.
-For example, when collecting nationwide weather temperature data,
-the city (e.g., `city="New York"`) is a typical tag column.
-In monitoring, system metrics like CPU and memory usually involve a `host` tag to represent the hostname.
+1. Defining the ordering of data in storage.
+   GreptimeDB reuses the `PRIMARY KEY` constraint to define tag columns and the ordering of tags.
+   Unlike traditional databases, in GreptimeDB, primary key and time series are identical.
```
killme2008, line 40:
```suggestion
   Unlike traditional databases, in GreptimeDB, the primary key and time series in typical TSDB are identical.
```

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
-Understanding the concept of the timeline is key to designing table structures and optimizing query performance. Properly organizing Tag columns, Field columns, and the Time Index can create an efficient data model that meets business needs.
+- The `'append_mode'='true'` option creates the table as an append-only table.
+  This ensures a log doesn't override another one with the same timestamp.
+- The table sorts logs by time so it is efficient to search logs by time.
 
-## Primary Key and Index
+### When to use primary key
 
-In GreptimeDB, data is organized sequentially based on the `PRIMARY KEY` columns and deduplicated based on the combination of `PRIMARY KEY` and `TIME INDEX` values. GreptimeDB supports data updates by inserting new rows to overwrite existing rows with the same `PRIMARY KEY` and `TIME INDEX` values.
+In GreptimeDB, the primary key is identical to the time-series.
```
killme2008, line 105:
```suggestion
In GreptimeDB, the primary key is identical to the time-series in typical TSDB.
```

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
+  http_method STRING,
+  http_refer STRING,
+  user_agent STRING,
+  request_id STRING,
+  request STRING,
+  PRIMARY KEY(application),
+) with ('append_mode'='true');
+```
+
+The unique values of the primary key in the table can significantly affect performance.
```
killme2008, line 136:
```suggestion
The unique values (AKA cardinality) of the primary key in the table can significantly affect performance.
```

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
 
-Recommendations for Field columns:
+- Low cardinality columns that occur in `WHERE`/`GROUP BY`/`ORDER BY` frequently.
+  These columns usually remain constant.
+  For example, `namespace`, `cluster`, or an AWS `region`.
+- No need to set all low cardinality columns as tags since this may impact the performance of ingestion and querying.
+- Typically use short strings for tags, avoiding `FLOAT`, `DOUBLE`, `TIMESTAMP`.
+- Never set high cardinality columns as tags if they change frequently.
+  For example, `trace_id`, `span_id`, `user_id` must not be used as tags.
+  GreptimeDB works well if you set them as fields instead of tags.
```
killme2008, line 74: Note: Creating indexes on fields and tags can be done at any time.

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
-By using the `INVERTED INDEX` column constraint,
-an inverted index is created specifically for the `idc` column.
-This approach helps to avoid potential performance and storage issues that could arise from the high cardinality of the `host` column.
 
-## Full-Text Index
+### Skipping Index
 
-For log text-type Field columns that require tokenization and inverted index-based querying, GreptimeDB provides full-text indexing.
+For high cardinality columns like `trace_id`, `request_id`, using a [skipping index](/user-guide/manage-data/data-index.md#skipping-index) is more appropriate.
+This method has lower storage overhead and resource usage, particularly in terms of memory consumption.
```
killme2008, line 173:
```suggestion
This method has lower storage overhead and resource usage, particularly in terms of memory and disk consumption.
```

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
+  request_id STRING,
+  request STRING,
+  PRIMARY KEY(application),
+) with ('append_mode'='true');
+```
 
-Example:
+The following query can use the inverted index on the `http_method` column.
 
 ```sql
```
killme2008, line 165: Do we need to explain which type of queries can benefit from indexes? Only the exact equal or prefix matching or range queries like that.

killme2008 (commented):
Great work! However, I think we should include some practical guidelines, particularly a series of 'how-to' guides. For instance, we could cover topics such as:

* Best practices for designing a traces table
* Guidelines for structuring a log table
* How to desgin table for prometheus-like metrics
* Guidelines for Iot Devices sensors
* And similar common use cases

docs/user-guide/concepts/data-model.md
```diff
@@ -7,31 +7,31 @@ description: Describes the data model of GreptimeDB, focusing on time-series tab
...
+GreptimeDB stores and processes data by time-series.
+Choosing the right schema is crucial for efficient data storage and retrieval; please refer to the [schema design guide](/docs/user-guide/administration/design-table.md) for more details.
 
-Suppose we have a time-series table called `system_metrics` that monitors the resource usage of a standalone device:
+### Metrics
+
+Suppose we have a table called `system_metrics` that monitors the resource usage of machines in data centers:
 
 ```sql
 CREATE TABLE IF NOT EXISTS system_metrics (
```
nicecui, line 37: Is this SQL statement still suitable for the new data model where the tag columns are no longer recommended?

docs/user-guide/administration/design-table.md
sunng87: > primary key and time series are identical

We can improve this by time series is defined by primary key

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
-In monitoring, system metrics like CPU and memory usually involve a `host` tag to represent the hostname.
+1. Defining the ordering of data in storage.
+   GreptimeDB reuses the `PRIMARY KEY` constraint to define tag columns and the ordering of tags.
+   Unlike traditional databases, in GreptimeDB, primary key and time series are identical.
+   Tables in GreptimeDB sort rows by (primary key, timestamp).
+   This improves the locality of data with the same tags.
+   If there are no tag columns, GreptimeDB sorts rows by timestamp.
+2. Identifying a unique time-series.
+   When the table is not append-only, GreptimeDB can deduplicate rows by timestamp under the same time-series (primary key).
+3. Smoothing migration from other TSDBs that use tags or labels.
```
sunng87, line 46: We need to mention the impact of the order of tags in primary key

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
 
-1. Storing low-cardinality metadata.
-2. Filtering data, such as using the `city` column to view the average temperature in New York City over the past week. This is similar to the `WHERE` clause in SQL.
-3. Grouping and aggregating data. For instance, if the temperature data includes a `state` label in addition to `city`, you can group the data by `state` and calculate the average temperature for each `state` over the past week. This is similar to the `GROUP BY` clause in SQL.
+### Choosing tag columns
 
-Recommendations for Tag columns:
+Usually you don't need tag columns since ordering data by timestamp is sufficient for most use cases.
+If you need deduplication, or your queries can benefit from the ordering, you can define tag columns.
+Setting tags may also reduce disk space usage as it improves the locality of data.
```
sunng87, line 53: I feel this is due to the limitation caused cardinality effect on memtable. This should not be, at least the final, definition of tags. 

The selection of tags determines:

1. how we can dedup data
2. how we can sort data to narrow down the scan range of user's most typical queries
3. how we can do effective partition, for small query there should be less partitions envolved.
4. sorted data brings better compression as mentioned

docs/user-guide/administration/design-table.md
sunng87: Some content for reference: https://clickhouse.com/docs/use-cases/observability/schema-design#choosing-a-primary-ordering-key

docs/user-guide/administration/design-table.md
```diff
@@ -19,220 +19,257 @@ Before proceeding, please review the GreptimeDB [Data Model Documentation](/user
...
-the city (e.g., `city="New York"`) is a typical tag column.
-In monitoring, system metrics like CPU and memory usually involve a `host` tag to represent the hostname.
+1. Defining the ordering of data in storage.
+   GreptimeDB reuses the `PRIMARY KEY` constraint to define tag columns and the ordering of tags.
+   Unlike traditional databases, in GreptimeDB, primary key and time series are identical.
+   Tables in GreptimeDB sort rows by (primary key, timestamp).
+   This improves the locality of data with the same tags.
+   If there are no tag columns, GreptimeDB sorts rows by timestamp.
+2. Identifying a unique time-series.
+   When the table is not append-only, GreptimeDB can deduplicate rows by timestamp under the same time-series (primary key).
```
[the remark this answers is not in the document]
evenyag: Append mode is intended not to dedup, some data like logs can have duplicate keys.

docs/user-guide/administration/design-table.md
evenyag: Removing the limitation may be the final goal, but this document is a guide for the current version.

However, a high cardinality primary key may still cause issues in some time-series workloads, like fetching last rows, processing in per-series manner. This may be related to the schema design.

docs/user-guide/administration/design-table.md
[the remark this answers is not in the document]
sunng87: I think we should at least keep a semantic for deduplicate entries. For performance consideration, we can do best-effort dedup, for example dedup on read or on compaction. Otherwise we will be in a position that in theory we cannot deal with logs generated from upstream retries.

docs/user-guide/concepts/data-model.md
evenyag: Tags are still helpful for querying if we set them properly. We can only use host for tag. But the problem is that I can't modify the SVG......

docs/user-guide/administration/design-table.md
evenyag: Currently, this is an implementation limitation. Our primary key is a time-series and is similar to a stream in VL.
https://docs.victoriametrics.com/victorialogs/keyconcepts/#how-to-determine-which-fields-must-be-associated-with-log-streams

In the future, we can support flexible primary key, but some optimization is only possible and available when the cardinality is low,

For `trace_id+spanid`, it is actually not a time-series. We may need to consider this in the future.

docs/user-guide/administration/design-table.md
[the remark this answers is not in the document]
evenyag: In most analytical databases, the primary key is not unique.

docs/user-guide/administration/design-table.md
evenyag: Not only memtable, our batch struct is also per series.
https://github.com/GreptimeTeam/greptimedb/blob/c77ce958a3b936e1065abe51fa10832b0ea2ac99/src/mito2/src/read.rs#L63-L74

docs/user-guide/administration/design-table.md
evenyag: The `When to use primary key` section explains this and has an example.

killme2008 (approved):
Let's go! It's a great step!

docs/user-guide/concepts/data-model.md
evenyag, line 1: I'll update the zh version in this PR.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/administration/design-table.md
```diff
@@ -16,284 +16,351 @@ description: 详细介绍了 GreptimeDB 的数据模型使用指南，以及常
...
-通常来说，为了达到最好的效果，我们推荐分区键尽量均匀分散且稳定，这通常需要一些关于数据分布方式的先验知识。如：
+- 本地磁盘空间不足以存储数据或在使用对象存储时缓存数据。
+- 你需要更多 CPU 来提高查询性能或更多内存用于高开销的查询。
+- 磁盘吞吐量成为瓶颈。
+- 写入速率大于单个节点的吞吐量。
 
-- 通过 MAC 地址的前/后缀来分区
-- 通过机房编号来分区
-- 通过业务名称
+GreptimeDB 在每次主要版本更新时都会发布[benchmark report](https://github.com/GreptimeTeam/greptimedb/tree/main/docs/benchmarks/tsbs)，
```
nicecui, line 345:
```suggestion
GreptimeDB 在每次主要版本更新时都会发布[benchmark report](https://github.com/GreptimeTeam/greptimedb/tree/VAR::greptimedbVersion/docs/benchmarks/tsbs)，
```

docs/user-guide/administration/design-table.md
```diff
@@ -288,17 +340,31 @@ with each partition containing all columns.
...
+
+There is no need to partition a table unless a single machine isn't enough to serve the table.
+For example:
+
+- There is not enough local disk space to store the data or to cache the data when using object stores.
+- You need more CPU cores to improve the query performance or more memory for costly queries.
+- The disk throughput becomes the bottleneck.
+- The ingestion rate is larger than the throughput of a single node.
+
 GreptimeDB releases a [benchmark report](https://github.com/GreptimeTeam/greptimedb/tree/main/docs/benchmarks/tsbs) with each major version update,
```
nicecui, line 358:
```suggestion
GreptimeDB releases a [benchmark report](https://github.com/GreptimeTeam/greptimedb/tree/VAR::greptimedbVersion/docs/benchmarks/tsbs) with each major version update,
```

nicecui (approved):

PR #1734, opened 2025-05-13
docs: add traces to the overview of user guide

nicecui:
## What's Changed in this PR

- Rename `traces` to `traces` at the sidebar
- add traces to the overview of user guide

## Checklist

- [x] Please ensure that the content in `sidebars.ts` matches the current document structure when you changed the document structure.


docs/user-guide/overview.md
```diff
@@ -31,6 +31,10 @@ To use GreptimeDB as a log storage solution,
 refer to the [Logs Documentation](./logs/overview.md).
 It details how to ingest pattern text logs using pipelines.
 
+To use GreptimeDB as a trace storage solution,
```
killme2008, line 34: What about using a list form:

* For logs user....
* For traces users ...

killme2008 (approved):
LGTM

docs/user-guide/overview.md
nicecui: According to the developer documentation style guide:
> address the reader of your documents using the second person instead of the first person
> Use the word `user` only to refer to the user of the software that your reader is developing.

If you want to let the `log/trace` appear at the beginning of the sentences, we can change it to:
- For log storage solution...
- For trace storage solution...

killme2008: I meant it to be a markdown list, but that's ok.

sunng87 (approved):

PR #1788, opened 2025-05-28
chore: add more details about altering ttl

### LINKED ISSUE #1781: explain more about altering TTL

nicecui:
## What's Changed in this PR

close #1781 

## Checklist


fengjiachun (approved):
LGTM

docs/reference/sql/alter.md
```diff
@@ -26,7 +26,10 @@ ALTER DATABASE db
 ```
 
 Currently following options are supported:
-- `ttl`: the default retention time of data in database.
+- `ttl`: Specifies the default retention time for data in the database, which is applied immediately upon modification.
+   - If `ttl` was not previously set, defining a new `ttl` using `ALTER` will result in the deletion of data that exceeds the specified retention time.
+   - If `ttl` was already set, modifying it via `ALTER` will enforce the updated retention time immediately, removing data that exceeds the new retention threshold.
```
killme2008, line 31: Data removal occurs in the background and will not take effect immediately.

shuiyisong (approved):

docs/reference/sql/alter.md
killme2008: Yes, but you have to explain why it doesn't take effect after alerting. Our users are developers, they will understand what we explain.

PR #1801, opened 2025-05-30
docs: add docs for configuring metasrv backend storage by helm chart

zyy17:
## What's Changed in this PR

Add docs for configuring metasrv backend storage by helm chart

## Checklist

- [x] Please ensure that the content in `sidebars.ts` matches the current document structure when you changed the document structure.
- [x] This change requires follow-up update in localized docs.

### DIFF
--- a/docs/user-guide/deployments/deploy-on-kubernetes/common-helm-chart-configurations.md
+++ b/docs/user-guide/deployments/deploy-on-kubernetes/common-helm-chart-configurations.md
@@ -430,3 +430,87 @@ debugPod:
       memory: 256Mi
       cpu: 200m
 ```
+
+### Configuring Metasrv Backend Storage
+
+#### Using MySQL and PostgreSQL as Backend Storage
+
+You can configure the backend storage for the metasrv through the `meta.backendStorage` field.
+
+Let's take MySQL as an example.
+
+```yaml
+meta:
+  backendStorage:
+    mysql:
+      # -- MySQL host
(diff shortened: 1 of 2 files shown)


daviderli614 (approved):

docs/user-guide/deployments/deploy-on-kubernetes/common-helm-chart-configurations.md
```diff
@@ -430,3 +430,87 @@ debugPod:
...
+The etcd backend storage can be configured through the `meta.backendStorage.etcd` field.
+
+```yaml
+meta:
+  backendStorage:
+    etcd:
+      # -- Etcd endpoints
+      etcdEndpoints: "etcd.etcd-cluster.svc.cluster.local:2379"
+      # -- Etcd store key prefix
+      storeKeyPrefix: ""
```
fengjiachun, line 512: All backend storage can have a `storeKeyPrefix`; we might need to add this to the `MySql` and `PostgreSQL` YAML examples.

WenyXu: I don’t think let's user to configure `storeKeyPrefix` with MySQL and PostgreSQL is a best practice. Instead, users should configure the `meta_table_name` directly. For PostgreSQL, users should also be able to configure the `meta_election_lock_id`. However, this change requires releasing a new version of the Operator. How about creating an issue to track this and moving forward with the implementation?

PR #1821, opened 2025-06-09
docs: remove to_unixtime references

masonyc:
## What's Changed in this PR

DataFusion already supports `to_unixtime` function, so remove the custom to_unixtime doc https://github.com/GreptimeTeam/greptimedb/pull/6271

## Checklist


killme2008:
As said in https://github.com/GreptimeTeam/greptimedb/pull/6271, we have to keep the current implementation due to the time zone issue.

PR #1848, opened 2025-06-19
fix: move headings out from template

sunchanglong:
## What's Changed in this PR

*Describe the change in this PR*

## Checklist


nicecui:
thanks to @sunchanglong , got the answer, but now we do not plan to move all headings out from template.

PR #2075, opened 2025-08-31
chore: update bitnami config

daviderli614:
## What's Changed in this PR

related to: https://github.com/bitnami/containers/issues/83267
Sync image to our repository.

## Checklist


sunng87:
Is there any other best practice from the community?

nicecui (approved):

daviderli614:
> LGTM as a short term solution. But in long term we should switch to a public repo, which feels more reliable and secure.

We can use a specific version of the public image to sync to our image repository.

daviderli614:
Now use the greptime dockerhub image.

PR #2134, opened 2025-09-19
fix: authentication helm chart value example

pbzweihander:
## What's Changed in this PR

According to the greptimedb-cluster Helm chart code, name of the user in auth block should be `username`, not `name`.

https://github.com/GreptimeTeam/helm-charts/blob/1843f2a53f02c0a4ffbf65b470996bf9a2bce404/charts/greptimedb-cluster/templates/users-auth-secret.yaml#L13

## Checklist


killme2008:
Thanks!

PR #2189, opened 2025-10-31
docs: refine user guide configuration

fengjiachun:
## What's Changed in this PR

As the title

## Checklist


docs/user-guide/deployments-administration/configuration.md
```diff
@@ -134,15 +134,29 @@ Note that HTTP and gRPC protocols cannot be disabled for the database to functio
 addr = "127.0.0.1:4000"
 timeout = "30s"
 body_limit = "64MB"
+#+ max_total_body_memory = "0"
+enable_cors = true
+prom_validation_mode = "strict"
 
 [grpc]
 bind_addr = "127.0.0.1:4001"
```
killme2008: I think `bind_addr` is outdated, and possibly other configurations as well. We need to check which items still exist in the documentation. And the `server_addr` is not described here.

PR #2453, opened 2026-05-04
docs: note that append_mode hint is ignored for InfluxDB auto-created tables

v0y4g3r:
## Summary

- Add a note to the InfluxDB line protocol docs that the `append_mode` HTTP header hint is ignored for auto-created tables, since `merge_mode` is always set to `last_non_null`.
- Update all versioned English docs (0.12–1.0), nightly English docs, and all Chinese docs (nightly + 0.12–1.0).

## Test plan

- [ ] Verify the new note renders correctly in the English nightly docs
- [ ] Verify the new note renders correctly in the Chinese nightly docs
- [ ] Confirm all versioned docs contain the same addition


v0y4g3r:
We need to change the behavior to changing merge mode to last_row when user explicitly set `append_mode` to true.

Issue #2417, opened 2026-04-15
Audit v1.0 documents


killme2008:
We need to audit all v1.0 documents, either using AI or manually, as some information is outdated.

* [x] Quick start  
* [x] User guide  
* [x] References  
* [x] Configuration  @WenyXu 
* [ ] Admin & deployment  
    * [x] Kubernetes & monitoring  @daviderli614 
    * [x] Manage metadata @WenyXu 
    * [ ] Trouble shooting
    * [ ] Manage data
    * [ ] Disaster Recovery 
    * [ ] .....

PR #2539, opened 2026-06-09
docs: fix contributor guide frontend docs

fengjiachun:
Part of #2417.

This PR covers the Contributor Guide / Frontend portion of the docs audit.

Changes:
- Update the frontend table sharding page in current and v1.0 docs, plus zh translations, to avoid the outdated claim that metasrv automatically moves regions or splits/merges them by data volume or access pattern. The page now points readers to explicit repartitioning for split/merge operations.
- Fix the zh distributed querying SQL example in current and v1.0 docs by adding the required time index, using the partition column `a` in partition expressions, and updating the follow-up query to reference an existing column.

Validation:
- Ran `git diff --check`.
- Verified the fixed distributed-querying SQL example against a local GreptimeDB MySQL endpoint at `127.0.0.1:4002`.
- Verified the previous example fails because it lacks a time index, and still fails with `Column "n" in rule expr is not referenced in PARTITION ON` when only the time index is added.


fengjiachun:
@codex review

killme2008 (approved):

PR #2544, opened 2026-06-09
docs: clarify deprecated flow streaming mode

discord9:
## Summary
- clarify that Flow batching mode remains active while the original streaming mode is deprecated
- update Flow/Flownode docs, glossary, and skills to avoid promoting streaming-mode wording
- document that ttl='instant' Flow sources currently fall back to deprecated streaming mode

## Validation
- git diff --check
- pnpm build

## Notes
- Historical release notes, versioned docs, and i18n docs are intentionally unchanged.


fengjiachun (approved):
LGTM

PR #2566, opened 2026-06-15
docs: document flow incremental read option

discord9:
## Summary
- Document the experimental `experimental_enable_incremental_read` Flow option.
- Explain append-only source table requirements and batching SQL limitations.
- Note full-snapshot/fallback behavior and `ttl=instant` limitation.

## Notes
- Current English docs only; translations/versioned docs can follow after approval.


docs/user-guide/flow-computation/manage-flow.md
```diff
@@ -101,6 +101,7 @@ CREATE [ OR REPLACE ] FLOW [ IF NOT EXISTS ] <flow-name>
 SINK TO <sink-table-name>
 [ EXPIRE AFTER <expr> ]
 [ COMMENT '<string>' ]
+[ WITH (<flow-option> = <value> [, ...]) ]
```
fengjiachun, line 104: Since this updates the CREATE FLOW grammar, could we also update `docs/reference/sql/create.md`? The SQL reference still shows CREATE FLOW without the `WITH (...)` flow-options clause, so users who check the reference will not see that `experimental_enable_incremental_read` is valid syntax.

fengjiachun (approved):
Others LGTM

PR #2592, opened 2026-06-30
docs: add enterprise dashboard webhook trigger docs

v0y4g3r:
## Summary

- Add a standalone Enterprise Management Console page for webhook triggers based on cluster resource usage metrics.
- Document configuration, supported metrics and roles, firing/resolved payload behavior, and in-memory state limitations.
- Add sidebar and console overview links, plus matching version `1.1` and Chinese translation pages.

## Availability

This feature is enterprise-only and available when the enterprise dashboard is deployed with dashboard version `v0.2.0-alpha.10` or later.

## Test Plan

- `node --input-type=module` MDX compilation for changed and new docs pages
- `git diff --check`
- `pnpm build`

### DIFF
--- a/docs/enterprise/console-ui/webhook-triggers.md
+++ b/docs/enterprise/console-ui/webhook-triggers.md
@@ -1,0 +1,85 @@
+---
+keywords: [enterprise, management console, webhook triggers, resource usage, alerting, CPU, memory]
+description: Configure enterprise dashboard webhook triggers to send notifications based on GreptimeDB cluster resource usage metrics.
+---
+
+# Webhook Triggers
+
+Webhook triggers monitor cluster resource usage metrics and send HTTP notifications when a configured threshold is reached. This is an enterprise-only feature and is available only when the enterprise dashboard is deployed.
+
+Configure webhook triggers per provisioned instance under `settings.monitoring.webhook_triggers` in the dashboard apiserver configuration. Enabled webhook triggers require a metrics source, either `settings.monitoring.greptimedb.url` or `settings.monitoring.metrics.prometheus`.
+
+```yaml
+provisionedInstances:
+  - name: mycluster
+    settings:
+      monitoring:
+        greptimedb:
+          url: http://monitoring-greptimedb:4000
+        # Or use a Prometheus-compatible metrics source:
+        # metrics:
+        #   prometheus: http://prometheus:9090
+        webhook_triggers:
+          - name: high-datanode-memory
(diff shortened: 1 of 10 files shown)


versioned_docs/version-1.1/enterprise/console-ui.md
```diff
@@ -59,4 +59,4 @@ Displays long-running SQL and PromQL queries with detailed execution time and qu
 
 ## Continuous Profiling
 
-The Management Console supports [Continuous Profiling](./continuous-profiling) for capturing and analyzing GreptimeDB component CPU and memory profiles.
+The Management Console supports [Continuous Profiling](./continuous-profiling) for capturing and analyzing GreptimeDB component CPU and memory profiles. It also supports [Webhook Triggers](./console-ui/webhook-triggers.md) for sending notifications based on cluster resource usage metrics.
```
discord9: Minor: the adjacent Continuous Profiling link still uses `./continuous-profiling`, which resolves to `/enterprise/continuous-profiling` from this page. Please change it to `./console-ui/continuous-profiling.md` here and in the v1.1 zh mirror.

docs/enterprise/console-ui/webhook-triggers.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+Webhook trigger configuration items:
+
+- `name`: Trigger name. Required when `enabled` is `true`. The name must be unique within the instance and must not contain `/`.
+- `enabled`: Enables or disables the trigger.
+- `roles`: Optional role filter. Omit it or leave it empty to match all roles. Supported roles are `frontend`, `metasrv`, `datanode`, and `flownode`.
+- `metric`: Resource usage metric to evaluate. Supported metrics are `memory_usage_percent`, `memory_usage_bytes`, `cpu_usage_percent`, and `cpu_usage_millicores`.
+- `operator`: Comparison operator. The default and only supported value is `>=`.
+- `threshold`: Threshold value. It must be greater than `0`. Percentage metrics must be less than or equal to `100`.
```
discord9, line 49: Nit: could we spell out the threshold units for non-percent metrics? `memory_usage_bytes` uses bytes and `cpu_usage_millicores` uses millicores, so values like `threshold: 1000` are otherwise a bit ambiguous. Please also apply the same clarification to the mirrored zh/versioned copies.

discord9 (approved):
LGTM. I left two minor follow-up comments; neither blocks this PR.

PR #2601, opened 2026-07-03
docs: document auto repartition column hint

WenyXu:
## What's Changed in this PR

This PR documents the GreptimeDB Enterprise Auto Repartition column hint for unpartitioned tables.

Changes include:

- Update Enterprise Auto Repartition docs to explain how unpartitioned tables can use `repartition.column.hint`.
- Clarify that the hint must be set through `ALTER TABLE ... SET`, and that `CREATE TABLE ... WITH` is not supported.
- Document hint restrictions:
  - only one column is supported;
  - the column must exist;
  - the column cannot be the time index column;
  - the table must not already have partition metadata;
  - the hint must be set or unset separately from other table options.
- Update the Autopilot overview limitation text.
- Add SQL Reference coverage for `ALTER TABLE ... SET/UNSET 'repartition.column.hint'`, marked as an Enterprise feature and linked to Auto Repartition.
- Add corresponding Chinese localized docs.

## Checklist

### DIFF
--- a/docs/enterprise/autopilot/auto-repartition.md
+++ b/docs/enterprise/autopilot/auto-repartition.md
@@ -31,9 +31,34 @@ Auto Repartition is useful in the following scenarios:
 - You want to split large Regions into smaller Regions and distribute potential bottlenecks through later scheduling.
 - You want to reduce the operational cost of manually identifying large Regions and running Repartition.
 
+## Auto Repartition for unpartitioned tables
+
+Auto Repartition can also work with unpartitioned tables when a repartition column hint is specified. For an unpartitioned table, GreptimeDB Enterprise does not infer partition columns automatically. You must first set the preferred column for future Auto Repartition by using `ALTER TABLE`:
+
+```sql
+ALTER TABLE sensor_readings SET 'repartition.column.hint' = 'host';
+```
+
+To remove the hint:
+
+```sql
+ALTER TABLE sensor_readings UNSET 'repartition.column.hint';
+```
+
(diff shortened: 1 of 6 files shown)


docs/enterprise/autopilot/auto-repartition.md
```diff
@@ -31,9 +31,34 @@ Auto Repartition is useful in the following scenarios:
...
+
+```sql
+ALTER TABLE sensor_readings UNSET 'repartition.column.hint';
+```
+
+The hint only records metadata for future Auto Repartition. It does not trigger Repartition immediately. After the table meets the Auto Repartition trigger conditions, GreptimeDB Enterprise can use the hinted column to generate partition boundaries and submit a Repartition action.
+
+The repartition column hint has the following restrictions:
+
+- It must be set with `ALTER TABLE ... SET`; setting it in `CREATE TABLE ... WITH` is not supported.
```
fengjiachun: The implementation already accepts this option from `CREATE TABLE ... WITH` too: `create_table_info()` validates `repartition.column.hint` during create, and sqlness covers `CREATE TABLE ... WITH ('repartition.column.hint' = 'host')` in `tests/cases/standalone/common/show/show_create.sql`. This restriction would incorrectly tell users a valid way to set the hint is unsupported. The zh mirror has the same sentence.

fengjiachun (commented):
Found one docs accuracy issue.

docs/enterprise/autopilot/auto-repartition.md
WenyXu: Good catch

WenyXu: [ec0363d](https://github.com/GreptimeTeam/docs/pull/2601/commits/ec0363d70682f02dc5bfb57552a9763018b66507)

fengjiachun (approved):
LGTM

PR #2609, opened 2026-07-06
docs: fix ingest and protocol details, sync EN/ZH

killme2008:
Fixes several inaccuracies in the ingest-data and protocol docs, and syncs the same changes to the Chinese nightly docs. Verified against GreptimeDB source and a live v1.1.1 instance; the OTel Collector config was validated with `otelcol validate`.

- **InfluxDB**: V2 write examples use `bucket`; document `db` as a V2 compatibility alias (verified in `influxdb_write_v2` and the auth path).
- **Elasticsearch**: `_bulk` uses `application/x-ndjson` + `--data-binary`; disable ES template/ILM/data stream in the Logstash/Filebeat examples; clarify that omitting `pipeline_name` falls back to `greptime_identity`.
- **Fluent Bit**: document the `custom_time_index` parameter; fix the OTLP API anchor links to point at the metrics/logs/traces sections.
- **Kafka**: note how Vector's `influxdb` codec maps metric names and the `val` value column.
- **OpenTelemetry**: correct the metrics exporter package names (were pointing at trace exporters); drop the stale `v0.16` wording.
- **OTel Collector**: keep the correct `otlphttp` exporter type (an earlier edit had broken it to `otlp_http`); reword to `OTLP/HTTP`.
- **Vector**: clarify the Loki sink writes to the default `loki_logs` table since it cannot set headers.
- **PostgreSQL / HTTP**: minor connect/param wording fixes.


shuiyisong (approved):
LGTM

PR #2639, opened 2026-07-20
docs: add per-region write buffer docs (`write_buffer_size`, `default_region_write_buffer_size`)

evenyag:
### Motivation
- Document a new per-region write-buffer feature so operators understand how to limit memtable usage per region and protect hot regions.
- Clarify precedence and interaction between table-level limits and the Mito engine defaults to avoid operator misconfiguration.

### Description
- Add a `write_buffer_size` table option to `CREATE TABLE` and `ALTER TABLE` docs with `SET`/`UNSET` examples.
- Document the configured value as the per-region stall threshold: GreptimeDB schedules a flush at half the value, stalls writes at the value, and rejects writes at twice the value.
- Distinguish an explicit table value of `0`, which disables the per-region limit, from `UNSET`, which falls back to `default_region_write_buffer_size`.
- Add `default_region_write_buffer_size` to the `region_engine.mito` example and option table, and update the performance-tuning guidance for hot-region protection.
- Apply these changes only to the nightly English and Chinese docs. The feature is not available in v1.1, so `versioned_docs/version-1.1` remains unchanged.

### Testing
- Ran `git diff --check` successfully.
- Ran `typos` on all eight modified Markdown files successfully.
- Verified there is no explicit-zero SQL example and the existing `UNSET 'write_buffer_size'` example remains.
- Verified that `versioned_docs/version-1.1` has no changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5dc5595bd08329b4f10ab6910a0a81)


docs/reference/sql/create.md
```diff
@@ -41,7 +41,7 @@ Database options behave differently:
 - **TTL and Compaction options** (`ttl`, `compaction.*`): These options have ongoing effect. Tables without specified values will continuously inherit database-level values. Changing the database TTL or compaction options will immediately impact all tables that don't have their own settings.
 
 - **Other options** (`memtable.type`, `append_mode`, `merge_mode`, `skip_wal`, `sst_format`): These act as template variables that are only applied when creating new tables. Changing these database-level options will NOT affect existing tables - they only serve as defaults for newly created tables.
-:::
+  :::
```
evenyag: Should we remove the whitespace?

docs/user-guide/deployments-administration/configuration.md
```diff
@@ -509,6 +510,7 @@ Available options:
...
 | `global_write_buffer_reject_size`        | String  | `2GB`         | Global write buffer size threshold to reject write requests. If not set, it's default to 2 times of `global_write_buffer_size`                                                                                                                                                                                                                                                                                                                                                                     |
+| `default_region_write_buffer_size`       | String  | `0`           | Default write buffer size for each region. Regions flush when they reach this size and reject writes at twice this size. Setting it to `0` disables the default per-region limit unless a table specifies `write_buffer_size`.                                                                                                                                                                                                                                                                     |
```
fengjiachun: The configured value is the stall threshold, not the flush threshold. `region_write_buffer_status_from_usage()` schedules a flush once mutable usage reaches half this value, stalls at this value, and rejects at twice it. Could we align this wording and its mirrors with those semantics?

docs/reference/sql/alter.md
```diff
@@ -193,6 +193,7 @@ Currently following options are supported:
 - `compaction.twcs.max_output_file_size`: the maximum allowed output file size of TWCS compaction strategy.
 - `compaction.twcs.trigger_file_num`: the number of files in a specific time window to trigger a compaction.
 - `sst_format`: the SST format of the table. The value can be `flat` or `primary_key`. A table supports changing the format in both directions: `primary_key` to `flat` and `flat` to `primary_key`.
+- `write_buffer_size`: the per-region write buffer limit of the table. Set it to a size such as `512MB` to make each region flush at that size and reject writes at twice that size. Set it to `0` or unset it to disable the table-level limit and use `region_engine.mito.default_region_write_buffer_size` if configured.
```
fengjiachun: `0` and `UNSET` have different fallback behavior: an explicit `0` disables the per-region limit even when the engine default is nonzero, while `UNSET` removes the table override and falls back to `default_region_write_buffer_size`. Could we document these cases separately here and in the mirrors?

docs/user-guide/deployments-administration/configuration.md
evenyag: Updated all nightly English and Chinese mirrors. The docs now describe the configured value as the stall threshold, with flush scheduled at half the value based on mutable memtable usage, stalling at the value, and rejection at twice the value.

docs/reference/sql/alter.md
evenyag: Updated all nightly English and Chinese mirrors. The docs now distinguish an explicit table value of 0, which disables the per-region limit even with a nonzero engine default, from UNSET, which removes the table override and falls back to the engine default. The SQL examples keep only the common UNSET case.

fengjiachun (commented):
En part LGTM

v0y4g3r (approved):

PR #2646, opened 2026-07-24
chore: add skill to convert influxql to greptimedb sql

MichaelScofield:
## What's Changed in this PR

as title

## Checklist


MichaelScofield:
@codex review

skills/README.md
```diff
@@ -61,3 +62,7 @@ Using `skills` cli tool to install the skill to your coding agents.
 ### `greptimedb-cluster-health-check`
 
 `npx skills add https://github.com/GreptimeTeam/docs/tree/main/skills/greptimedb-cluster-health-check`
+
+### `influxql-to-greptimedb-sql`
```
killme2008, line 66: Link the skill to the doc  https://docs.greptime.com/user-guide/migrate-to-greptimedb/migrate-from-influxdb/ too

skills/influxql-to-greptimedb-sql/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+  max(usage_system) RANGE '10m' FILL PREV AS max_usage_system
+FROM cpu
+WHERE regexp_like(host, '^web-')
+  AND ts >= now() - '6 hours'::INTERVAL
+ALIGN '10m' BY (host)
+ORDER BY ts DESC
+LIMIT 100;
+```
+
+### Sources
```
killme2008, line 335: Add a link  https://docs.greptime.com/user-guide/migrate-to-greptimedb/migrate-from-influxdb/

skills/influxql-to-greptimedb-sql/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+
+- GreptimeDB Range Query: <https://docs.greptime.com/reference/sql/range/>
+- GreptimeDB SQL functions: <https://docs.greptime.com/reference/sql/functions/df-functions/>
```
killme2008: Better to link the overview https://docs.greptime.com/reference/sql/functions/overview/

skills/influxql-to-greptimedb-sql/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+  AND ts < now();
+```
+
+Preserve open and closed bounds. Do not change `<` to `<=`.
+
+InfluxQL `GROUP BY time()` queries without an explicit time predicate stop at `now()` by default. Add `<time_index> <= now()` when exact behavior matters.
+
+#### `GROUP BY time()`
+
+InfluxQL defaults an omitted `fill()` to `fill(null)`. Use Range Query with `FILL NULL` to preserve empty windows:
```
killme2008: `FILL NULL` does not preserve empty windows the way InfluxQL `fill(null)` does. Tested on 1.1.3: `FILL` only fills gaps between each series' first and last data-bearing windows. It never emits windows before the first or after the last data point, and the window range differs per series. InfluxQL emits every window in the query time bound, for every series.

Repro: data at 10:10–10:31, `WHERE ts >= '10:00' AND ts < '11:00'`, `RANGE '5m' FILL NULL ALIGN '5m' BY (host)` — output starts at 10:10 and ends at 10:30; a series with a single point yields a single row. InfluxQL would return 10:00 through 10:55 for all series. After a Grafana migration this shows up as missing leading/trailing buckets.

Suggest rewording "preserve empty windows" and adding this to the mandatory `Semantic differences` list — by this skill's own rule, the conversion must not be presented as exact.

skills/influxql-to-greptimedb-sql/SKILL.md
```diff
@@ -0,0 +1,304 @@
...
+  ts,
+  host,
+  avg(usage_user) RANGE '10m' FILL PREV AS mean_usage_user,
+  max(usage_system) RANGE '10m' FILL PREV AS max_usage_system
+FROM cpu
+WHERE regexp_like(host, '^web-')
+  AND ts >= now() - '6 hours'::INTERVAL
+ALIGN '10m' BY (host)
+ORDER BY ts DESC
+LIMIT 100;
```
killme2008, line 323: This converts InfluxQL's per-series `LIMIT 100` into a global SQL `LIMIT 100` with no note — the silent approximation the skill itself forbids ("never describe an approximate conversion as exact"), and the `fill(previous)` window-boundary difference from the comment above applies here too. The example is the strongest behavioral signal in a skill; suggest appending the `Assumptions` / `Semantic differences` blocks defined in the Output section so it demonstrates the full expected format.

skills/influxql-to-greptimedb-sql/SKILL.md
```diff
@@ -0,0 +1,304 @@
+---
+name: influxql-to-greptimedb-sql
+description: Convert InfluxQL SELECT queries into SQL supported by GreptimeDB and identify semantics that cannot be translated exactly. Use for InfluxDB-to-GreptimeDB migrations, Grafana query rewrites, InfluxQL time-window and fill conversions, function migrations, or requests mentioning "InfluxQL to SQL", "convert InfluxQL", "migrate InfluxDB queries", or "rewrite InfluxQL".
```
killme2008, line 3: The description advertises "Grafana query rewrites" but the body has nothing Grafana-specific. Migrated InfluxQL panels are almost always `WHERE $timeFilter GROUP BY time($__interval)`. A short section mapping `$timeFilter` → `$__timeFilter(ts)` and `time($__interval)` → `RANGE '$__interval' ... ALIGN '$__interval'` would make that advertised use case actually work.

killme2008 (commented):
Ran every SQL snippet in this skill against a local GreptimeDB 1.1.3 — all of them work as written, including the `ALIGN TO` no-cast caveat (`TO '...'::TIMESTAMP` does fail with `Unsupported keyword: ::`). Nice work on the mappings. Three issues below, first two are correctness.

killme2008:
Please add the skill to

* https://docs.greptime.com/faq-and-others/vibecoding/
* https://docs.greptime.com/user-guide/migrate-to-greptimedb/migrate-from-influxdb/

skills/influxql-to-greptimedb-sql/SKILL.md
killme2008: And https://docs.greptime.com/user-guide/query-data/sql/

MichaelScofield:
@killme2008 PTAL

killme2008 (approved):
LGTM

Please backport them into v1.1 @MichaelScofield

PR #2654, opened 2026-07-27
docs: add Query Guard plugin docs for banning DROP TABLE/DATABASE

v0y4g3r:
## What's changed

Add documentation for the GreptimeDB Enterprise **Query Guard** plugin, including the new `ban_drop_table` / `ban_drop_database` options introduced in [GreptimeTeam/greptimedb-enterprise#832](https://github.com/GreptimeTeam/greptimedb-enterprise/pull/832).

- **New page** `enterprise/deployments-administration/query-guard` (EN + ZH):
  - How the bans work: enforced in query interceptors **before permission checks**, so they apply to all users including admins
  - SQL path (MySQL/PostgreSQL/HTTP) rejects `DROP TABLE` / `DROP DATABASE`; gRPC path rejects `DROP TABLE` DDL requests (no drop-database variant in gRPC)
  - Internal operations (TTL, auto-cleanup) are unaffected
  - Configuration example: `[[plugins]] [plugins.query_guard]` with `enable`, `ban_drop_table`, `ban_drop_database`
  - Caveats: every frontend must carry the config in multi-frontend deployments; enabling the plugin also turns on the pre-existing `disallow_cross_catalog_query` behavior (documented here for the first time)
- **Sidebar**: registered under Enterprise → Deployments & Administration
- **Cross-links**: added a tip in the `DROP` SQL reference (EN + ZH) pointing to the new page

## Verification

- `pnpm typecheck` ✅
- `pnpm build` (EN) ✅
- `DOC_LANG=zh pnpm build` (ZH) ✅

### DIFF
--- a/docs/enterprise/deployments-administration/query-guard.md
+++ b/docs/enterprise/deployments-administration/query-guard.md
@@ -11,8 +11,10 @@ It provides the following protections:
 
 - **Ban `DROP TABLE`**: reject all `DROP TABLE` statements.
 - **Ban `DROP DATABASE`**: reject all `DROP DATABASE` statements.
-- **Disallow cross-catalog queries**: reject queries that reference tables across
-  different catalogs.
+- **Reject `COPY` statements**: reject all `COPY` statements.
+- **Disallow cross-catalog access**: reject queries that reference tables across
+  different catalogs, cross-catalog gRPC DDL requests, and Flight bulk inserts
+  targeting a different catalog.
 
 ## Overview
 
@@ -58,8 +60,10 @@ it takes effect on the frontend where it is configured.
 - **Every frontend must carry the configuration.** In a deployment with multiple
   frontends, you must add the plugin configuration to every frontend's config file;
   frontends without the configuration will still execute `DROP` statements.
-- **Enabling the plugin also disallows cross-catalog queries.** Turning on
-  `query_guard` automatically enables `disallow_cross_catalog_query`, even if you
-  only want the `DROP` bans.
+- **Enabling the plugin activates all protections except the `DROP` bans.**
+  Turning on `query_guard` automatically rejects `COPY` statements, cross-catalog
+  queries, cross-catalog gRPC DDL requests, and Flight bulk inserts targeting a
+  different catalog, even if you only want the `DROP` bans. The `DROP` bans are
+  controlled separately by `ban_drop_table` and `ban_drop_database`.
 - The bans are enforced at the protocol layer. Changing the configuration requires
   a restart of the frontend to take effect.
--- a/docs/enterprise/deployments-administration/query-guard.md
+++ b/docs/enterprise/deployments-administration/query-guard.md
@@ -29,8 +29,10 @@ protocol:
 
 - SQL path: `DROP TABLE` and `DROP DATABASE` statements are rejected with a
   `NotSupported` error.
-- gRPC path: `DROP TABLE` DDL requests are rejected. Note that `DROP DATABASE` is
-  SQL-only; the gRPC protocol has no drop-database request.
+- gRPC path: structured `DROP TABLE` DDL requests are rejected. The structured
+  gRPC DDL request has no drop-database variant; however, SQL statements sent over
+  gRPC go through the same SQL interceptors, so the `DROP DATABASE` ban applies to
+  SQL over gRPC as well.
 
 Internal operations such as TTL-based data expiration and automatic cleanup bypass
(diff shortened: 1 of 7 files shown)


shuiyisong (commented):
Two source-validated documentation findings.

docs/enterprise/deployments-administration/query-guard.md
```diff
@@ -0,0 +1,65 @@
+---
+keywords: [query guard, ban drop table, ban drop database, data protection, security, configuration]
+description: Guide to configuring the Query Guard plugin in GreptimeDB Enterprise to ban DROP TABLE / DROP DATABASE statements for all users and disallow cross-catalog queries.
+---
+
+# Query Guard
+
+Query Guard is a GreptimeDB Enterprise plugin that intercepts queries at the frontend
+protocol layer and rejects potentially dangerous statements before they are executed.
+It provides the following protections:
```
shuiyisong, line 10: **[P2] Document all restrictions activated by Query Guard.**

This supposedly complete list omits behavior enabled regardless of the two `DROP` flags. The SQL interceptor always falls through to [`reject_copy`](https://github.com/GreptimeTeam/greptimedb-enterprise/blob/4dac9e3f2175d18afe6ad471d9c30e909b5fbcb0/src/ent-plugins/src/frontend/query_guard.rs#L59-L70), so every `COPY` variant is rejected. The same guard also rejects explicit cross-catalog gRPC DDL and Flight bulk inserts through [`validate_ddl_expr` and `pre_bulk_insert`](https://github.com/GreptimeTeam/greptimedb-enterprise/blob/4dac9e3f2175d18afe6ad471d9c30e909b5fbcb0/src/ent-plugins/src/frontend/query_guard.rs#L90-L195), while this page only mentions cross-catalog queries. An operator enabling Query Guard solely for the `DROP` bans can therefore break existing `COPY`, DDL, or Flight workflows. Please document these always-on effects here or under Caveats, and mirror the change in the Chinese page.

docs/enterprise/deployments-administration/query-guard.md
```diff
@@ -0,0 +1,65 @@
...
+to **every user, including administrators**. No one can drop tables or databases
+through the client protocols until the configuration is changed and the frontend
+is restarted.
+
+The bans cover both the SQL protocols (MySQL, PostgreSQL, and HTTP) and the gRPC
+protocol:
+
+- SQL path: `DROP TABLE` and `DROP DATABASE` statements are rejected with a
+  `NotSupported` error.
+- gRPC path: `DROP TABLE` DDL requests are rejected. Note that `DROP DATABASE` is
```
shuiyisong: **[P2] Qualify the gRPC `DROP DATABASE` statement.**

This is true only for the structured `DdlExpr` request, whose enum has no `DropDatabase` variant. The gRPC API also accepts [`Request::Query(Query::Sql)`](https://github.com/GreptimeTeam/greptimedb/blob/f5f5d468bbc8fa68e65b46e8541a3370c68b9838/src/client/src/database.rs#L679-L688), and the frontend [routes that request through `do_query_inner`](https://github.com/GreptimeTeam/greptimedb/blob/f5f5d468bbc8fa68e65b46e8541a3370c68b9838/src/frontend/src/instance/grpc.rs#L101-L116), where the SQL Query Guard rejects `DROP DATABASE`. Please say that the *structured gRPC DDL request* lacks a `DropDatabase` variant, while SQL sent over gRPC is subject to both bans; mirror the clarification in the Chinese page.

docs/enterprise/deployments-administration/query-guard.md
v0y4g3r: Fixed in 7e31d308. The protections list now documents the always-on restrictions: COPY statements are rejected, and cross-catalog access rejection now covers cross-catalog gRPC DDL requests and Flight bulk inserts targeting a different catalog. The Caveats section was also updated to state that enabling `query_guard` activates all protections except the DROP bans, which remain controlled separately by `ban_drop_table` / `ban_drop_database`. Mirrored in the Chinese page.

docs/enterprise/deployments-administration/query-guard.md
v0y4g3r: Fixed in cb43a7c2. The gRPC bullet now clarifies that only the structured gRPC DDL request lacks a drop-database variant, while SQL statements sent over gRPC pass through the same SQL interceptors, so the `DROP DATABASE` ban applies to SQL over gRPC as well. Mirrored in the Chinese page.

shuiyisong (approved):

PR #2668, opened 2026-07-30
docs: add multi-tenancy FAQ entry

killme2008:
Add a "How do I implement multi-tenancy?" FAQ entry to the Deployment & Operations section, recommending isolation by database or table dimension and noting that a large number of tables is fine given GreptimeDB's horizontally scalable architecture.

Applied to EN and ZH, current and v1.1.

Ref: https://github.com/GreptimeTeam/greptimedb/issues/8689

### DIFF
--- a/docs/faq-and-others/faq.md
+++ b/docs/faq-and-others/faq.md
@@ -312,12 +312,12 @@ ORDER BY files DESC;
 
 The recommended approach is to isolate tenants at the **database** or **table** level, rather than mixing every tenant into one shared table distinguished only by a tag.
 
-- **Database per tenant**: Give each tenant its own [database](/reference/sql/create.md#create-database). This provides the cleanest isolation — separate schemas, per-tenant TTL and other table defaults, and independent access control.
+- **Database per tenant**: Give each tenant its own [database](/reference/sql/create.md#create-database). This keeps each tenant's data cleanly separated, with independent schema, TTL, and other table defaults.
 - **Table per tenant**: Within a shared database, give each tenant its own table (or table prefix). A good fit when tenants share configuration but you still want data-level separation.
 
 Don't worry about ending up with a large number of databases or tables. GreptimeDB's distributed architecture scales horizontally — tables are split into regions spread across datanodes, and you add datanodes as the table and region count grows. There is no hard limit on the number of tables, and performance depends on primary key design rather than table count (see [scalability characteristics](#what-are-greptimedbs-scalability-characteristics)).
 
-For per-tenant access control, use [user-level authentication](/user-guide/deployments-administration/authentication/overview.md) in the open-source version, or RBAC in Enterprise / GreptimeCloud.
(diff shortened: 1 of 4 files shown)


docs/faq-and-others/faq.md
```diff
@@ -308,6 +308,17 @@ ORDER BY files DESC;
...
+
+- **Database per tenant**: Give each tenant its own [database](/reference/sql/create.md#create-database). This provides the cleanest isolation — separate schemas, per-tenant TTL and other table defaults, and independent access control.
+- **Table per tenant**: Within a shared database, give each tenant its own table (or table prefix). A good fit when tenants share configuration but you still want data-level separation.
+
+Don't worry about ending up with a large number of databases or tables. GreptimeDB's distributed architecture scales horizontally — tables are split into regions spread across datanodes, and you add datanodes as the table and region count grows. There is no hard limit on the number of tables, and performance depends on primary key design rather than table count (see [scalability characteristics](#what-are-greptimedbs-scalability-characteristics)).
+
+For per-tenant access control, use [user-level authentication](/user-guide/deployments-administration/authentication/overview.md) in the open-source version, or RBAC in Enterprise / GreptimeCloud.
```
fengjiachun: OSS authentication does not provide per-tenant database or table isolation: the static user provider authorizes every catalog and schema, and the default permission checker only distinguishes read versus write without checking table targets. A readonly user can therefore still query every tenant database or table. Please limit this recommendation, and the independent access control claim above, to Enterprise or Cloud RBAC and ACL; otherwise document that OSS requires separate deployments or an external enforcement layer.

docs/faq-and-others/faq.md
```diff
@@ -308,6 +308,17 @@ ORDER BY files DESC;
...
 
+### How do I implement multi-tenancy?
+
+The recommended approach is to isolate tenants at the **database** or **table** level, rather than mixing every tenant into one shared table distinguished only by a tag.
+
+- **Database per tenant**: Give each tenant its own [database](/reference/sql/create.md#create-database). This provides the cleanest isolation — separate schemas, per-tenant TTL and other table defaults, and independent access control.
+- **Table per tenant**: Within a shared database, give each tenant its own table (or table prefix). A good fit when tenants share configuration but you still want data-level separation.
+
+Don't worry about ending up with a large number of databases or tables. GreptimeDB's distributed architecture scales horizontally — tables are split into regions spread across datanodes, and you add datanodes as the table and region count grows. There is no hard limit on the number of tables, and performance depends on primary key design rather than table count (see [scalability characteristics](#what-are-greptimedbs-scalability-characteristics)).
```
fengjiachun, line 318: This makes the table count sound cost-free, but the table-design and Prometheus guides both note that each physical table maintains its own files and that many small tables can hurt storage and query efficiency. Horizontal scaling does not remove that overhead. Could we say there is no fixed hard limit while still recommending capacity planning for many tenant tables?

docs/faq-and-others/faq.md
killme2008: Yes, the user have to design the per-tenant authentication in their applications. I'll note it.

docs/faq-and-others/faq.md
killme2008: I don’t think it’s necessary; users always need to do capacity planning.

docs/faq-and-others/faq.md
killme2008: Fixed in https://github.com/GreptimeTeam/docs/pull/2668/changes/30924d8e8060c874f8a3ff1033a9b2a9743b19d6

fengjiachun (approved):
LGTM

sunng87 (approved):

PR #2730, opened 2026-08-10
docs: document streaming EXPLAIN ANALYZE HTTP API

### LINKED ISSUE #2719: Document the streaming EXPLAIN ANALYZE HTTP API
## Context

GreptimeDB v1.2 introduces a streaming SQL analysis endpoint:

- Upstream: https://github.com/GreptimeTeam/greptimedb/pull/8380
- Endpoint: `POST /v1/sql/analyze/stream`
- Transport: Server-Sent Events (SSE)

The current docs only expose the related configuration and do not document the API contract.

## Documentation work

Add HTTP API/reference documentation covering:

- A request example using one `EXPLAIN ANALYZE VERBOSE` statement.
- Accepted format constraints and rejection of non-explain or multi-statement requests.
- SSE event types: `metrics`, `final`, `canceled`, and `error`.
- Metrics as complete, best-effort snapshots rather than deltas, including snapshot coalescing.
- Terminal event behavior and client disconnect cancellation.
- No resume, reconnect, or detached-query lifecycle.
- The v1.2 default for the related configuration, verified against the v1.2 source/tag.

## Scope

Update both English and Chinese documentation in:

- Nightly
- v1.2

## Completion criteria

- [ ] The endpoint has a complete request and event-stream example.
- [ ] Lifecycle, cancellation, and snapshot semantics are explicit.
- [ ] Version-sensitive defaults are verified against v1.2.
- [ ] English and Chinese content is aligned.
- [ ] Affected locale builds pass.

discord9:
## What changed

Document the `POST /v1/sql/analyze/stream` (SSE) endpoint contract added in v1.2 (upstream GreptimeDB #8380): request parameters, `EXPLAIN ANALYZE VERBOSE` statement restrictions, the `metrics`/`final`/`canceled`/`error` SSE events and payload fields, snapshot semantics (complete best-effort snapshot with adaptive coalescing), client-disconnect cancellation, and the no resume/reconnect lifecycle.

Closes #2719.

## Scope

- Documentation versions: Nightly, 1.2
- Languages: English, Chinese

## Verification

- Contract verified against GreptimeDB v1.2 source (src/servers/src/http/handler.rs SSE event generation; endpoint registration in src/servers/src/http.rs; config default `http.experimental_enable_explain_analyze_stream = true`, `snapshot_interval_ms` default 5000 clamped to [1000, 60000]).
- EN/ZH and Nightly/v1.2 copies are line-aligned.
- markdownlint config in repo disables all rules; manual structural checks passed. Docusaurus build not run (no node runtime available).

## Checklist


docs/reference/http-endpoints.md
```diff
@@ -215,6 +215,80 @@ Various query APIs for sending query to GreptimeDB.
...
+| `sql` | Yes | The `EXPLAIN ANALYZE VERBOSE` statement to execute. |
+| `db` | No | The database to run the statement against. |
+| `snapshot_interval_ms` | No | Interval between `metrics` snapshots in milliseconds. Defaults to `5000`; values are clamped to the `[1000, 60000]` range. |
+
+Example:
+
+```bash
+curl -N -X POST 'http://127.0.0.1:4000/v1/sql/analyze/stream' \
+  -H 'Accept: text/event-stream' \
+  -F 'sql=EXPLAIN ANALYZE VERBOSE SELECT * FROM monitor'
```
fengjiachun, line 243: The example uses `curl -F`, which sends `multipart/form-data`, but this handler extracts Axum `Form<SqlQuery>`, which accepts `application/x-www-form-urlencoded`. Because the unsupported media type is ignored and no query-string `sql` is supplied, this command reaches `sql parameter is required` instead of opening an SSE stream. Please use `--data-urlencode 'sql=EXPLAIN ANALYZE VERBOSE SELECT * FROM monitor'` and update all four copies so the documented example works.

discord9: Fixed in 70c4f17e. All four examples (English/Chinese, Nightly/v1.2) now use `--data-urlencode`, and the request parameter text explicitly specifies `application/x-www-form-urlencoded`. I verified the corrected curl command with a local HTTP request capture: both the Content-Type and decoded SQL body are correct. The merge conflicts are also resolved while preserving the upstream Format SQL and Parse SQL sections.

docs/reference/http-endpoints.md
```diff
@@ -219,6 +219,78 @@ These endpoints execute SQL, PromQL, or structured log queries.
...
+The endpoint accepts one SQL `EXPLAIN ANALYZE VERBOSE` statement or one `TQL ANALYZE VERBOSE` statement. A `FORMAT JSON` clause is optional; no other output format is supported. Any other request is rejected with a regular JSON error response (not SSE), including:
+
+- Non-analyze statements, such as `SELECT ...`
+- SQL `EXPLAIN` or `EXPLAIN ANALYZE` without `VERBOSE`
+- An analyze statement with a non-JSON format, such as `FORMAT TEXT` or `FORMAT GRAPHVIZ`
+- Multiple statements separated by semicolons
+
+#### SSE events
+
+Each event is sent as an `event:` line followed by a `data:` line containing a JSON payload, and events are separated by blank lines. A keep-alive comment line is sent every 15 seconds while the stream is open.
```
killme2008, line 259: axum's `KeepAlive` is an idle timeout, not a fixed tick — the timer is reset on every emitted event (`KeepAliveStream::poll_next` in `axum-0.8/src/response/sse.rs`). With the default 5000 ms snapshot interval a keep-alive comment is never sent at all. As written, a client would size its read timeout around a 15s heartbeat that does not exist.

Suggest: "A keep-alive comment line is sent after 15 seconds without any event."

docs/reference/http-endpoints.md
```diff
@@ -219,6 +219,78 @@ These endpoints execute SQL, PromQL, or structured log queries.
...
+Payload fields:
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `seq` | integer | Monotonically increasing sequence number of the event. |
+| `state` | string | The event type: `metrics`, `final`, `canceled`, or `error`. |
+| `partial` | boolean | `true` for `metrics` events, `false` for terminal events. |
+| `elapsed_ms` | integer | Elapsed time in milliseconds since the request started. |
+| `metrics` | array | When available, a `stage` / `node` / `plan` snapshot. It can appear in `metrics`, `final`, `canceled`, and `error` events. Periodic `metrics` events contain compact execution topology and runtime metrics; terminal events include verbose plan details. |
+| `output` | object | The final query result in GreptimeDB JSON format. Only present in `final` events. |
```
killme2008, line 270: `output` is the result of the `EXPLAIN ANALYZE` statement itself — the `stage`/`node`/`plan` records plus the total-rows row assembled in `create_output_batch` (`src/query/src/analyze.rs`) — identical to what `/v1/sql` returns for `EXPLAIN ANALYZE VERBOSE FORMAT JSON`. It does not contain the rows of the query being analyzed.

"The final query result" reads like the `SELECT * FROM monitor` rows, so a client written against this description will parse for something that never arrives. Worth stating explicitly.

i18n/zh/docusaurus-plugin-content-docs/current/reference/http-endpoints.md
```diff
@@ -219,6 +219,78 @@ curl -X POST 'http://127.0.0.1:4000/debug/prof/mem/gdump' -d 'activate=true'
...
+该端点接受一条 SQL `EXPLAIN ANALYZE VERBOSE` 语句或一条 `TQL ANALYZE VERBOSE` 语句。`FORMAT JSON` 子句可选，不支持其他输出格式。其他请求都会被拒绝，并返回常规的 JSON 错误响应（而非 SSE），包括：
+
+- 非 analyze 语句，例如 `SELECT ...`
+- 不带 `VERBOSE` 的 SQL `EXPLAIN` 或 `EXPLAIN ANALYZE`
+- 使用非 JSON 格式的 analyze 语句，如 `FORMAT TEXT` 或 `FORMAT GRAPHVIZ`
+- 以分号分隔的多条语句
+
+#### SSE 事件
+
+每个事件由 `event:` 行和包含 JSON 负载的 `data:` 行组成，事件之间以空行分隔。流打开期间，服务器每 15 秒发送一行 keep-alive 注释。
```
killme2008, line 259: 同 `docs/reference/http-endpoints.md:259`：keep-alive 是 idle timeout，不是固定 15 秒一次。

killme2008 (commented):
Checked the contract against main (nightly) and the `v1.2.0` tag. The version split is right: nightly dropped the config gate and gained TQL, 1.2 has the gate and SQL-only — the pages match. Interval/clamp constants, the adaptive 1 MiB/10 MiB thresholds, the `stage`/`node`/`plan` shape, `1005`, the `skip_serializing_if` behaviour of `output`/`reason`/`code`, and the nightly-only "slow clients may skip snapshots" note all check out.

Two things are wrong and worth fixing before merge (both repeated across all four files); inline.

Two optional follow-ups:

- Nightly only: "compact execution topology" overstates what `verbose=false` does. It only switches the per-node `param` string between `DisplayFormatType::Default` and `Verbose` (`one_line()` in `src/common/recordbatch/src/adapter.rs`). Node count, tree shape and the metric set are identical, and the stage-1 sub-stage metrics are unaffected by the flag. "compact" reads as "fewer nodes".
- Out of scope for this PR, but directly contradicts the new nightly page: `docs/user-guide/deployments-administration/configuration.md` still lists `experimental_enable_explain_analyze_stream` (example TOML and the options table). GreptimeDB #8966 removed the option from `HttpOptions` on main; only a `test_http_options_legacy_analyze_stream_config_is_ignored` test remains.

i18n/zh/docusaurus-plugin-content-docs/current/reference/http-endpoints.md
```diff
@@ -219,6 +219,78 @@ curl -X POST 'http://127.0.0.1:4000/debug/prof/mem/gdump' -d 'activate=true'
...
+负载字段：
+
+| 字段 | 类型 | 描述 |
+| --- | --- | --- |
+| `seq` | 整数 | 事件的单调递增序号。 |
+| `state` | 字符串 | 事件类型：`metrics`、`final`、`canceled` 或 `error`。 |
+| `partial` | 布尔值 | `metrics` 事件为 `true`，终止事件为 `false`。 |
+| `elapsed_ms` | 整数 | 自请求开始以来的耗时（毫秒）。 |
+| `metrics` | 数组 | 可用时，包含 `stage` / `node` / `plan` 快照。可以出现在 `metrics`、`final`、`canceled` 和 `error` 事件中。周期性 `metrics` 事件包含紧凑执行拓扑和运行时指标；终止事件包含详细计划。 |
+| `output` | 对象 | 以 GreptimeDB JSON 格式返回的最终查询结果。仅出现在 `final` 事件中。 |
```
killme2008, line 270: 同 `docs/reference/http-endpoints.md:270`：`output` 是 `EXPLAIN ANALYZE` 语句自身的结果表，不是被分析查询的数据行。

versioned_docs/version-1.2/reference/http-endpoints.md
```diff
@@ -219,6 +219,80 @@ These endpoints execute SQL, PromQL, or structured log queries.
...
+The endpoint only accepts a single `EXPLAIN ANALYZE VERBOSE` statement. `EXPLAIN ANALYZE VERBOSE FORMAT JSON` is accepted as well; the `FORMAT` clause is optional and only `JSON` is supported. Any other request is rejected with a regular JSON error response (not SSE), including:
+
+- Non-explain statements, such as `SELECT ...`
+- `EXPLAIN` or `EXPLAIN ANALYZE` without `VERBOSE`
+- `EXPLAIN ANALYZE VERBOSE` with a non-JSON format, such as `FORMAT TEXT` or `FORMAT GRAPHVIZ`
+- Multiple statements separated by semicolons
+
+#### SSE events
+
+Each event is sent as an `event:` line followed by a `data:` line containing a JSON payload, and events are separated by blank lines. A keep-alive comment line is sent every 15 seconds while the stream is open.
```
killme2008, line 261: Same as `docs/reference/http-endpoints.md:259` — the v1.2 handler uses the same `KeepAlive::new().interval(Duration::from_secs(15))`, so the idle-timeout wording applies here too.

versioned_docs/version-1.2/reference/http-endpoints.md
```diff
@@ -219,6 +219,80 @@ These endpoints execute SQL, PromQL, or structured log queries.
...
+Payload fields:
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `seq` | integer | Monotonically increasing sequence number of the event. |
+| `state` | string | The event type: `metrics`, `final`, `canceled`, or `error`. |
+| `partial` | boolean | `true` for `metrics` events, `false` for terminal events. |
+| `elapsed_ms` | integer | Elapsed time in milliseconds since the request started. |
+| `metrics` | array | When available, the current `EXPLAIN ANALYZE VERBOSE` metrics snapshot: an array of `stage` / `node` / `plan` entries. It can appear in `metrics`, `final`, `canceled`, and `error` events. |
+| `output` | object | The final query result in GreptimeDB JSON format. Only present in `final` events. |
```
killme2008, line 272: Same as `docs/reference/http-endpoints.md:270`.

i18n/zh/docusaurus-plugin-content-docs/version-1.2/reference/http-endpoints.md
```diff
@@ -219,6 +219,80 @@ curl -X POST 'http://127.0.0.1:4000/debug/prof/mem/gdump' -d 'activate=true'
...
+该端点仅接受单条 `EXPLAIN ANALYZE VERBOSE` 语句，也接受 `EXPLAIN ANALYZE VERBOSE FORMAT JSON`；`FORMAT` 子句可选，且仅支持 `JSON`。其他请求都会被拒绝，并返回常规的 JSON 错误响应（而非 SSE），包括：
+
+- 非 explain 语句，如 `SELECT ...`
+- 不带 `VERBOSE` 的 `EXPLAIN` 或 `EXPLAIN ANALYZE`
+- 使用非 JSON 格式的 `EXPLAIN ANALYZE VERBOSE`，如 `FORMAT TEXT` 或 `FORMAT GRAPHVIZ`
+- 以分号分隔的多条语句
+
+#### SSE 事件
+
+每个事件由 `event:` 行和包含 JSON 负载的 `data:` 行组成，事件之间以空行分隔。流打开期间，服务器每 15 秒发送一行 keep-alive 注释。
```
killme2008, line 261: 同 `docs/reference/http-endpoints.md:259`。

i18n/zh/docusaurus-plugin-content-docs/version-1.2/reference/http-endpoints.md
```diff
@@ -219,6 +219,80 @@ curl -X POST 'http://127.0.0.1:4000/debug/prof/mem/gdump' -d 'activate=true'
...
+负载字段：
+
+| 字段 | 类型 | 描述 |
+| --- | --- | --- |
+| `seq` | 整数 | 事件的单调递增序号。 |
+| `state` | 字符串 | 事件类型：`metrics`、`final`、`canceled` 或 `error`。 |
+| `partial` | 布尔值 | `metrics` 事件为 `true`，终止事件为 `false`。 |
+| `elapsed_ms` | 整数 | 自请求开始以来的耗时（毫秒）。 |
+| `metrics` | 数组 | 可用时，当前的 `EXPLAIN ANALYZE VERBOSE` 指标快照：由 `stage` / `node` / `plan` 条目组成的数组。可以出现在 `metrics`、`final`、`canceled` 和 `error` 事件中。 |
+| `output` | 对象 | 以 GreptimeDB JSON 格式返回的最终查询结果。仅出现在 `final` 事件中。 |
```
killme2008, line 272: 同 `docs/reference/http-endpoints.md:270`。

Issue #2723, opened 2026-08-10
Add an Events query guide


WenyXu:
### What documentation do you need?

Add an Events guide for the `greptime_private.events` system table.

Place it under **Deployments & Administration > Monitoring**, at the same navigation level as **Slow Query**.

The page should provide a query cookbook that helps users quickly inspect events. At minimum, include examples for:

```sql
-- Recent events
SELECT *
FROM greptime_private.events
ORDER BY timestamp DESC
LIMIT 20;
```

```sql
-- All lifecycle rows for one procedure
SELECT *
FROM greptime_private.events
WHERE procedure_id = '<procedure_id>'
ORDER BY timestamp ASC;
```

Also include examples for:

- Listing and filtering event types.
- Finding the latest event for a table, database, flow, view, or region.
- Querying failed procedure events.
- Reading `payload`, `event_context`, and `procedure_trigger`.
- Every supported event type or event family.

Link to the existing event recorder configuration instead of duplicating detailed configuration content.

### Who is this documentation for?

Users and operators who need to inspect DDL, procedure lifecycle, migration, repartition, GC, WAL, and other operational events without reading GreptimeDB source code.

### Versions and languages

Nightly and v1.2, English and Chinese

### Suggested content

Add an English and Chinese Events page and include it next to Slow Query in the corresponding navigation.

Validate event types, columns, and SQL examples against the GreptimeDB source code.

Tracked by #2715.

PR #2733, opened 2026-08-10
docs: add Events query and DDL guides

WenyXu:
## What changed

- add Nightly English and Chinese guides for querying `greptime_private.events`
- document Procedure lifecycles, event fields, JSON payloads, and event context
- document database, table, Flow, and View DDL event lifecycles
- add source-backed SQL examples for `Running`/`Submitted` and `Done`/`Succeeded` rows
- add Events navigation and reference links

Part of #2723

## Scope

- Documentation versions: Nightly only
- Languages: English and Chinese
- This PR covers the Events overview, query patterns, Procedure lifecycle, data model, and local DDL event families.
- Metasrv operational events are in a stacked follow-up PR.

## Verification

- `git diff --check`
- `pnpm build`
- `DOC_LANG=zh pnpm build`
- Manual source and SQL-evidence review

## Checklist

- [x] I updated the relevant documentation versions and languages.
- [x] I updated navigation.


WenyXu:
@codex review

docs/user-guide/deployments-administration/monitoring/events/query-events.md
```diff
@@ -0,0 +1,164 @@
...
+List types actually present in the cluster before choosing a filter:
+
+```sql
+SELECT type, COUNT(*) AS event_rows
+FROM greptime_private.events
+WHERE timestamp >= now() - INTERVAL '1' hour
+GROUP BY type
+ORDER BY type;
+```
+
```
evenyag, line 51: Are these docs too verbose? We have already document the schema of the table.

WenyXu:
@codex review

docs/user-guide/deployments-administration/monitoring/events/query-events.md
WenyXu: [ad3adfe](https://github.com/GreptimeTeam/docs/pull/2733/commits/ad3adfed9b828d65146f9bff48a6cc5cc8c8e52d)

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
```diff
@@ -0,0 +1,134 @@
...
+## Procedure event columns
+
+Procedure events also have the following columns:
+
+| Column              | Meaning                                                                                                                                                     |
+| ------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `procedure_id`      | Unique Procedure ID.                                                                                                                                        |
+| `procedure_state`   | Procedure state when the event was recorded. Values are `Running`, `Done`, `Retrying`, `PrepareRollback`, `RollingBack`, `Failed`, and `Poisoned`.          |
+| `procedure_trigger` | Procedure event trigger in JSON. Its `type` is `Submitted`, `Recovered`, `ChildSubmitted`, `Retrying`, `RollingBack`, `Succeeded`, `Failed`, or `Poisoned`. |
+| `procedure_error`   | Error message when the Procedure fails.                                                                                                                     |
```
fengjiachun: `procedure_error` is never SQL `NULL` — it is an empty string on success. From `src/common/procedure/src/event.rs`:

```rust
let error = match &self.state {
    ProcedureState::Failed { error }
    | ProcedureState::PrepareRollback { error }
    | ProcedureState::RollingBack { error }
    | ProcedureState::Retrying { error }
    | ProcedureState::Poisoned { error, .. } => format!("{error:?}"),
    _ => String::new(),
};
```

Describing it as "Error message when the Procedure fails" invites `WHERE procedure_error IS NOT NULL`, which matches every row. Worth stating that failures are selected with `procedure_error != ''`, or by filtering on `procedure_state` as the "Find failed Procedures" section already does. This is the easiest thing on these pages for a reader to get wrong.

Minor related point: the value is `format!("{error:?}")`, i.e. the `Debug` rendering of the error rather than a plain message.

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
```diff
@@ -0,0 +1,134 @@
...
+`Submitted` events normally have state `Running`. When a Procedure succeeds,
+the completed event has state `Done` and trigger type `Succeeded`. The completed
+row is generated from the Procedure's final state, so its fields can differ from
+the submitted row. Events are recorded asynchronously; a recording failure does
+not change the Procedure result.
+
+When `event_context` is available, its stable `reason` value is one of
+`manual`, `auto_create`, `auto_alter`, `auto_repartition`, `auto_rebalance`,
+`region_failover`, `scheduled_gc`, or `unknown`. For example, a MySQL-submitted
+event can contain `{"protocol":"mysql","reason":"manual"}`.
```
fengjiachun: Worth adding that `event_context` is written **only** on `Submitted` rows. From `src/common/procedure/src/event.rs`:

```rust
let event_context = matches!(self.trigger, EventTrigger::Submitted)
    .then_some(self.context.event_context.as_ref())
    .flatten()
```

This is what makes the `AND event_context IS NOT NULL` filter in the example below return exactly the submitted row. Documenting the rule both explains that example and gives readers a convenient way to select submitted rows.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Configure event recording
+
+The event recorder can be configured in standalone deployments or on Metasrv in distributed deployments. See [Event recorder](/user-guide/deployments-administration/configuration.md#lifecycle-event-recorder) for configuration options and the supported type list.
+
+Standalone deployments record supported local DDL Procedure events. Distributed deployments with Metasrv can additionally record operational event types.
+
+## Query events
+
+Start with a bounded query while investigating a recent operation:
```
fengjiachun, line 27: The events table is created lazily, so this first query is the one most likely to fail for a new reader.

There is no `CREATE TABLE` for `greptime_private.events` anywhere; the table is materialized by the first insert, with `ttl` and `append_mode` passed as hints (`EventHandlerOptions::to_hints` in `src/common/event-recorder/src/recorder.rs`). On a deployment where nothing has been recorded yet, or where `event_types = []` disables recording, this returns a table-not-found error rather than zero rows. One sentence here would save that surprise.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
```diff
@@ -0,0 +1,248 @@
...
+       region_migration_src_node_id, region_migration_dst_node_id
+FROM greptime_private.events
+WHERE type IN ('region_migration', 'batch_gc', 'repartition_group')
+  AND timestamp >= now() - INTERVAL '1' hour
+  AND (region_id = <region_id>
+       OR source_region_id = <region_id>
+       OR target_region_id = <region_id>)
+ORDER BY timestamp DESC
+LIMIT 1;
+```
```
fengjiachun: Two problems with this section.

First, columns only exist once that event family has actually been recorded. `build_row_inserts_request` builds the request schema from `event.extra_schema()` per event type, so the table grows incrementally and columns belonging to event types that never occurred are absent — naming them in a `SELECT` fails with a column-not-found error rather than returning `NULL`.

This query references three families at once (`region_id`, `source_region_id`/`target_region_id`, and the `region_migration_*` columns), so it only runs where `region_migration`, `repartition_group`, and `batch_gc` events have all been recorded.

Second, that puts the section outside this PR's own stated scope — the description scopes it to standalone and local DDL events, with Metasrv operational events deferred to the stacked follow-up. Suggest moving this section to that PR, or at minimum documenting that column availability depends on which event types have been recorded.

docs/user-guide/deployments-administration/monitoring/events/ddl-events.md
```diff
@@ -0,0 +1,213 @@
...
+This query returns table events of the selected type from the specified database
+in the last hour. To query one table, add
+`AND table_name = '<table_name>'` to the `WHERE` clause.
+
+Rows for the same table retain `table_name`. A successful `create_table` or
+`create_logical_tables` event includes `table_id` only when the Procedure returns
+the allocated ID. `alter_table`, `truncate_table`, and `drop_table` already know
+the table ID when they are submitted, so their event rows include it.
+
+GreptimeDB also records `create_logical_tables` and `alter_logical_tables` events.
```
fengjiachun, line 83: Two details about the logical-table families that readers will hit:

- `create_logical_tables` and `alter_logical_tables` emit **one row per logical table**, sharing a `procedure_id` and timestamp (the event envelope iterates over locators). Someone querying by `procedure_id` in the "Query a Procedure" section will see N rows per lifecycle stage, which is currently unexplained.
- The `alter_logical_tables` terminal row carries a `NULL` `table_id` (the integration test passes `None`), unlike `alter_table`. The paragraph above groups the alter family together, so this exception is easy to miss.

docs/user-guide/deployments-administration/monitoring/events/ddl-events.md
```diff
@@ -0,0 +1,213 @@
...
+`undrop_table` and `purge_dropped_table` are Enterprise-only.
```
fengjiachun: This is correct — both variants are behind `#[cfg(feature = "enterprise")]` in `src/common/meta/src/ddl/event/table.rs`.

It does contradict `configuration.md`, though, which lists `undrop_table` and `purge_dropped_table` unannotated in both the standalone and Metasrv supported-type blocks — and the Events overview points readers at exactly that list for "the supported type list". Pre-existing issue, but this PR is a natural place to annotate it.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,34 @@
...
+
+The `greptime_private.events` table stores events recorded while GreptimeDB runs. It helps operators inspect DDL changes and background procedures without searching service logs.
+
+GreptimeDB writes event records asynchronously on a best-effort basis. A recent operation might not be visible immediately, and event recording should not be treated as the operation's success acknowledgement.
+
+## Configure event recording
+
+The event recorder can be configured in standalone deployments or on Metasrv in distributed deployments. See [Event recorder](/user-guide/deployments-administration/configuration.md#lifecycle-event-recorder) for configuration options and the supported type list.
+
+Standalone deployments record supported local DDL Procedure events. Distributed deployments with Metasrv can additionally record operational event types.
```
fengjiachun, line 19: `admin_function` events have since landed: `943eee852f` (#8835, 2026-08-11) added `ADMIN_FUNCTION_EVENT_TYPE = "admin_function"`. That is newer than this branch, so not a gap in this PR, but it is neither a DDL Procedure event nor a Metasrv operational event, so this sentence's two-way split will need revisiting — along with the `configuration.md` type lists. Either here or as a follow-up.

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+## Procedure event type-specific columns
+
+The following columns are populated only by the listed event types. A `NULL`
+value in one of these columns is normal when it does not apply.
+
+- **Database, table, Flow, and view events:** `catalog_name`, `schema_name`,
+  `table_name`, `table_id`, `physical_table_id`, `flow_name`, `flow_id`,
+  `view_name`, and `view_id` identify the affected object. See [DDL events](/user-guide/deployments-administration/monitoring/events/ddl-events.md).
```
fengjiachun: Two scoping details in this bullet:

- `physical_table_id` applies only to logical-table events — `has_physical_table_id()` matches `CreateLogicalTables | AlterLogicalTables` only, not the database, Flow, or view families listed here.
- Repartition events also carry `catalog_name`, `schema_name`, `table_name`, and `table_id` (`RepartitionEvent::base_schema`), which this grouping does not mention.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
```diff
@@ -0,0 +1,248 @@
...
+  AND timestamp >= now() - INTERVAL '1' hour
+  AND schema_name = '<database_name>'
+  AND table_name = '<table_name>'
+ORDER BY timestamp DESC
+LIMIT 1;
+```
+
+### Flow
+
+Flow rows can have SQL `NULL` in `schema_name`. Use `catalog_name` plus a unique
```
fengjiachun: "can have SQL `NULL`" understates it: `FlowDdlEvent` has no `schema_name` field at all, so the column is *always* `NULL` for Flow events. Stating that outright makes the advice to use `catalog_name` plus `flow_name` more obviously necessary rather than merely defensive.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,34 @@
+---
+keywords: [GreptimeDB events, Procedure events, operational events, greptime_private]
+description: Query event records in greptime_private.events.
+---
+
+# Events
+
+The `greptime_private.events` table stores events recorded while GreptimeDB runs. It helps operators inspect DDL changes and background procedures without searching service logs.
+
+GreptimeDB writes event records asynchronously on a best-effort basis. A recent operation might not be visible immediately, and event recording should not be treated as the operation's success acknowledgement.
```
fengjiachun: Could be quantified: events are flushed on a 5s interval (`DEFAULT_FLUSH_INTERVAL_SECONDS` in `src/common/event-recorder/src/recorder.rs`), so "might not be visible immediately" has a concrete bound worth giving the reader.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Configure event recording
+
+The event recorder can be configured in standalone deployments or on Metasrv in distributed deployments. See [Event recorder](/user-guide/deployments-administration/configuration.md#lifecycle-event-recorder) for configuration options and the supported type list.
```
fengjiachun: Link text is "Event recorder" but the target heading is "Lifecycle event recorder" (same mismatch in the Chinese page, which links to `#生命周期事件记录器`). Matching the heading makes it clearer which config section to look for.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,34 @@
...
+Standalone deployments record supported local DDL Procedure events. Distributed deployments with Metasrv can additionally record operational event types.
+
+## Query events
+
+Start with a bounded query while investigating a recent operation:
+
+```sql
+SELECT timestamp, type
+FROM greptime_private.events
+WHERE timestamp >= now() - INTERVAL '1' hour
```
fengjiachun, line 32: **Correction — disregard this comment; the revert in 5e060b2 is right.**

I checked the wrong precedent. `reference/sql/data-types.md` describes the dialect forms for a standalone interval literal, but for timestamp arithmetic this repo consistently uses `INTERVAL '<n>' <unit>`:

- `user-guide/flow-computation/continuous-aggregation.md` — `now() - INTERVAL '1' minute`, `INTERVAL '2' minute`, `INTERVAL '30' second`
- `user-guide/query-data/cte.md` — `now() - interval '1' hour`
- the alternative house form is the cast, `now() - '5 minutes'::INTERVAL` (`user-guide/query-data/sql.md`)

There are zero occurrences of `now() - INTERVAL '<n unit>'` anywhere under `docs/` or `versioned_docs/`, and the two `now() - INTERVAL` cases in the source's sqlness suite both use `'1' day`. The current state — 42 uniform `INTERVAL '1' hour` across these pages — is correct and consistent. Sorry for the round-trip.

docs/reference/sql/greptime-private/overview.md
```diff
@@ -11,5 +11,6 @@ GreptimeDB stores some important internal information as system tables in the `g
 
 | Table Name                          | Description                                                                                   |
 | ----------------------------------- | --------------------------------------------------------------------------------------------- |
+| [`events`](/user-guide/deployments-administration/monitoring/events/event-data-model.md) | Stores events recorded while GreptimeDB runs. |
```
fengjiachun: This row links out to a user-guide page while its two siblings link to reference pages in this same directory (`./slow_queries.md`, `./pipelines.md`). Not wrong, but a reader scanning the table gets sent somewhere structurally different. Worth considering whether `greptime_private.events` warrants a reference page of its own like the other two system tables.

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
WenyXu: Updated. The column description now clarifies that failed Procedures store the Debug-formatted error, while successful rows contain an empty string.

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
WenyXu: Updated. The data-model page now states that `event_context` is written only on `Submitted` rows, which also explains the example filter.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Added this note to the overview and the new system-table reference page: the table is created when the first event is recorded, and querying it before that—or with recording disabled—returns a table-not-found error.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
WenyXu: Kept the Region query because it is useful for operational investigation, and documented its prerequisite: every listed event type must have been recorded before its columns are available in the table schema.

docs/user-guide/deployments-administration/monitoring/events/ddl-events.md
WenyXu: Documented that `create_logical_tables` and `alter_logical_tables` emit one row per logical table for each lifecycle trigger, so one `procedure_id` can return multiple rows for the same trigger. We intentionally do not call out the unavailable logical-table `table_id` as user-facing behavior.

docs/user-guide/deployments-administration/monitoring/events/ddl-events.md
WenyXu: Updated both standalone and Metasrv type lists in `configuration.md` to mark `undrop_table` and `purge_dropped_table` as GreptimeDB Enterprise-only.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Thanks. `admin_function` is outside this PR’s scope, so I have left it for the follow-up documentation work rather than expanding the current Events DDL guide.

docs/user-guide/deployments-administration/monitoring/events/event-data-model.md
WenyXu: Updated the grouping description: `physical_table_id` is now scoped to logical-table events only, and the repartition entry now includes its catalog, schema, table name, and table ID locators.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
WenyXu: Updated. The guide now says that Flows do not have a schema, so `schema_name` is SQL `NULL`; use `catalog_name` and `flow_name` to identify a Flow.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Added the normal five-second flush interval to the overview, while keeping the asynchronous-write caveat.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Updated the link text to “Lifecycle event recorder” so it matches the destination heading.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Normalized and live-validated all documented queries with the MySQL-compatible form `INTERVAL '1' hour`. The alternative `INTERVAL '1 hour'` does not parse on the validated GreptimeDB MySQL endpoint.

docs/reference/sql/greptime-private/overview.md
WenyXu: Added a concise `greptime_private.events` reference page in both languages, linked the system-table overview to it, and added it to the SQL reference sidebar. The page points readers to the Events guide for the detailed data model and query scenarios.

docs/user-guide/deployments-administration/monitoring/events/overview.md
```diff
@@ -0,0 +1,41 @@
...
+
+Standalone deployments record supported local DDL Procedure events. Distributed deployments with Metasrv can additionally record operational event types.
+
+## Query events
+
+Start with a bounded query while investigating a recent operation:
+
+The table is created when the first event is recorded. If no event has been
+recorded yet, or event recording is disabled, this query returns a table-not-found
+error.
```
fengjiachun: The content is right, but it landed between the lead-in and the block it introduces: "Start with a bounded query while investigating a recent operation:" ends on a colon, and this paragraph now sits between that colon and the fence.

Suggest reordering so the lead-in stays adjacent to its code block — the table-not-found caveat first, then "Start with a bounded query while investigating a recent operation:", then the SQL block.

The Chinese page has the same problem at the corresponding lines: its lead-in `排查最近的操作时，可以先查询最近一小时的事件，并将结果限制为 20 条：` also ends on a colon and is now separated from its code block by the new paragraph.

docs/reference/sql/greptime-private/events.md
```diff
@@ -0,0 +1,26 @@
+---
+keywords: [events, greptime private, system tables]
+description: The events table in the `greptime_private` database.
+---
+
+# events
+
+The `events` table stores lifecycle and operational events recorded while
+GreptimeDB runs. Event recording is asynchronous and best-effort, so rows are
+not an acknowledgement that an operation succeeded.
```
fengjiachun: "lifecycle and operational events" is true of a distributed deployment but not a standalone one, which records only local DDL Procedure events. Since this reference page is the entry point from the `greptime_private` table list, it may be worth mirroring the split the user guide already makes rather than presenting both categories as unconditional.

docs/reference/sql/greptime-private/events.md
```diff
@@ -0,0 +1,26 @@
...
+USE greptime_private;
+
+SELECT timestamp, type
+FROM events
+WHERE timestamp >= now() - INTERVAL '1' hour
+ORDER BY timestamp DESC
+LIMIT 20;
+```
+
+For the event data model and queries for DDL Procedures, see [Events](/user-guide/deployments-administration/monitoring/events/overview.md).
```
fengjiachun: Structurally this page diverges from its two siblings in the same directory. `slow_queries.md` and `pipelines.md` both follow "sample query → sample output → per-column description list", so a reader who opens them from the sidebar gets the table's schema. This page has no column information at all and forwards to the user guide instead.

Either is defensible, but a reader who navigates to `reference/sql/greptime-private/events` specifically to look up a column currently has to leave the reference section to find one. Two options: add the column list here (or a compact subset with a pointer for the per-event-type columns), or make the hand-off explicit — "column definitions are documented in [Event data model]" rather than the current, more general "For the event data model and queries for DDL Procedures".

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md
```diff
@@ -535,6 +535,8 @@ drop_table, undrop_table, purge_dropped_table, truncate_table,
 create_view, drop_view
 ```
 
+以上列表中的 `undrop_table` 和 `purge_dropped_table` 仅 GreptimeDB 企业版支持。
```
fengjiachun: The two Enterprise notes are worded differently: this one is prefixed with 以上列表中的, while line 553 is not. The English page uses an identical sentence in both places.

Each note sits directly under its own code block, so the prefix does not disambiguate anything — suggest matching line 553 in both spots.

docs/user-guide/deployments-administration/monitoring/events/overview.md
WenyXu: Updated the English and Chinese overviews. The table-creation caveat now appears before the query lead-in, so the lead-in remains directly adjacent to its SQL block.

docs/reference/sql/greptime-private/events.md
WenyXu: Updated the reference page to distinguish deployment scope: standalone records supported local DDL Procedure events, while distributed deployments with Metasrv can additionally record operational event types.

docs/reference/sql/greptime-private/events.md
WenyXu: Updated the handoff to explicitly link to [Event data model] for column definitions, with a separate link to the Events guide for DDL Procedure query examples.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md
WenyXu: Updated the first Chinese note to match the second one. Both now use the same Enterprise-only wording.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
```diff
@@ -0,0 +1,240 @@
...
+
+```sql
+SELECT timestamp, type, catalog_name, schema_name,
+       flow_name, flow_id, procedure_state
+FROM greptime_private.events
+WHERE timestamp >= now() - INTERVAL '1' hour
+  AND flow_name = '<flow_name>'
+ORDER BY timestamp DESC
+LIMIT 1;
+```
```
fengjiachun, line 120: Dropping the `catalog_name` filter makes sense, but the same commit removed the sentence that explained this section, and the projection was left as-is. The query still selects `catalog_name` and `schema_name` while filtering on neither, and `schema_name` is permanently SQL `NULL` for Flow rows — a reader running this now sees a NULL column with nothing on the page to say that is expected.

The explanation also went away from the `event-data-model.md` bullet in the same commit (`Flows do not have a schema, so their `schema_name` is SQL `NULL``), so it now survives only on `ddl-events.md`. Both locales are affected.

Two ways out:

- Drop `catalog_name, schema_name` from the projection. Neither is filtered here, and the other three subsections do not project `catalog_name` — this would also make Flow consistent with them.
- Keep the projection and restore one sentence: Flows have no schema, so `schema_name` is SQL `NULL`; locate them by `flow_name`.

The first seems cleaner given the goal of the commit was simplification.

WenyXu: Updated in 45805929. I took the first option: the Flow query now projects only `flow_name`, `flow_id`, and `procedure_state`, so it no longer exposes the unfiltered `catalog_name` or permanently `NULL` `schema_name` columns. The Chinese example is aligned.

docs/user-guide/deployments-administration/monitoring/events/query-events.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Query Procedure events
+
+Procedure events share a `procedure_id`.
+
+### Get a procedure ID
+
+For a table-creation procedure, locate the `Submitted` row by its catalog,
```
fengjiachun: Leftover from the filter simplification: the query below no longer filters on `catalog_name`, so "by its catalog, database, and table" now describes three locators where only two are used. Suggest "by its database and table".

The Chinese page has the same leftover at the corresponding line: `对于创建表的 Procedure，可以根据 catalog、数据库、表和 `Submitted` 触发器定位提交行：`.

The related sentence further down was already updated in the same commit ("Filtering by `schema_name` and `table_name` avoids selecting a procedure for another object with a similar name"), so this one looks simply missed.

WenyXu: Updated in c9909b9a. The English and Chinese text now says to filter by database and table, then select the row whose `procedure_trigger.type` is `Submitted`, matching the query.

fengjiachun (approved):
LGTM

PR #2748, opened 2026-08-14
docs: iceberg metadata export

sunng87:
## What changed

Initial version of iceberg user guide in enterprise section.

## Scope

- Documentation versions: nightly, 1.2
- Languages: both

TODO: backport to 1.2 if everything looks good.

## Checklist

- [ ] I verified the content against the applicable GreptimeDB version.
- [ ] I updated the relevant documentation versions and languages, or explained why not.
- [ ] I checked changed links and anchors.


docs/enterprise/iceberg.md
```diff
@@ -0,0 +1,293 @@
...
+
+This is a **read-only export**, not a copy or a dual-write. GreptimeDB keeps writing data the way it always has;
+the Iceberg integration simply publishes the metadata that lets other engines find and interpret those same files.
+
+## How it works
+
+The key idea is that GreptimeDB already stores its SST (sorted string table) data files as Parquet in object storage.
+The Iceberg integration does **not** re-write, duplicate, or export the data. Instead it publishes Iceberg
+**metadata** — manifests, manifest-lists, and table-metadata snapshots — that point at the existing Parquet files.
+Any engine that speaks Iceberg can then read those files through the REST catalog.
```
killme2008, line 20: It would be better to include a diagram.

docs/enterprise/iceberg.md
```diff
@@ -0,0 +1,293 @@
...
+```
+http://<frontend-http-host>:<port>/v1/iceberg
+```
+
+For example, with a frontend listening on the default HTTP port `4000`, the catalog base URI is
+`http://localhost:4000/v1/iceberg` and the warehouse name is `greptime` (the default).
+
+## Make a table readable
+
+GreptimeDB publishes Iceberg metadata on flush, so the export of newly-written data appears after the next flush
```
killme2008: newly -> lately?

docs/enterprise/iceberg.md
```diff
@@ -0,0 +1,293 @@
...
+  snapshots or roll back.
+- **No full schema history.** The current schema is always exposed, but past schema versions are not retained, so
+  you cannot reconstruct what a table looked like at an earlier point in time.
+
+### Data types and Spark
+
+GreptimeDB types map cleanly to Iceberg for the common cases (booleans, signed integers, floats, strings, binary,
+date, timestamp). A few things to be aware of, especially in Spark:
+
+- **Declare the time index as `TIMESTAMP(6)`.** GreptimeDB's default `TIMESTAMP` is millisecond precision, but the
```
killme2008, line 279: In that case, does it mean historical tables cannot use this feature, and only newly created tables with the correct TIME INDEX type configured can use it? I think this issue needs to be addressed.

sunng87: this is hard to fix from iceberg side because only microsends is supported from iceberg, especially for spark implementation, it assumes stats data in parquet files are in microseconds unit.

I'm trying to allow user to run alter table to change the time index unit.

killme2008 (approved):

PR #2762, opened 2026-08-18
docs: split Kafka deployment and Remote WAL guidance

killme2008:
## What changed

`deploy-kafka.md` and `wal/remote-wal/manage-kafka.md` both covered Helm installation, verification and Remote WAL endpoints, but with different chart versions (31.0.0 vs 32.4.3), image tags (`-r1` vs `-r12`), namespaces and resource sizing. Neither page read as authoritative.

- `deploy-kafka.md` stays the installation guide (install, verify, monitoring, Grafana dashboard, uninstall).
- `manage-kafka.md` becomes the Remote WAL perspective: what Kafka holds in this setup, and links to deployment, requirements and GreptimeDB-side configuration. It no longer repeats the install steps.
- Removed `log.retention.bytes=250000000` from the example `extraConfig`, and annotated `log.retention.hours=4` as a testing value. `wal/remote-wal/configuration.md` already states that Remote WAL topics must not use size-based retention; Kafka deletes a segment as soon as any retention condition matches, which can remove WAL that GreptimeDB has not replayed. `overwrite_entry_start_id = true` does not protect against this — it advances the read position to the earliest available offset and skips the deleted entries.
- Fixed a duplicated `kafka-controller-0` in the expected `kubectl get pod` output.

No version numbers were changed. Chart 31.0.0 has appVersion 3.9.0, which matches the image `deploy-kafka.md` pins. The mismatched pairing was `manage-kafka.md`'s chart 32.4.3 (appVersion 4.0.0) with a 3.9.0 image; it goes away with the install section.

Both URLs stay live — the repository has no client-redirects plugin, so neither page could be deleted.

## Scope

- Languages: English, Chinese

## Verification

```
git diff --check
DOC_LANG=en pnpm check:links
DOC_LANG=zh pnpm check:links
```

Both link checks pass. `onBrokenAnchors` is `throw` under `DOCS_LINK_CHECK`, so the two new cross-page anchors are validated. Also confirmed against the build output that both `deploy-kafka` and `manage-kafka` still emit `index.html`, and that the Chinese `configuration` page really carries `id="注意事项与限制"`.

## Known item

`manage-kafka.md` previously set `global.security.allowInsecureImages: true`; `deploy-kafka.md` does not. Chart 31.0.0 (released 2024-11-12) predates the Bitnami image guard, so it likely does not need the key — but `deploy-kafka.md`'s expected output shows `Substituted images detected`, which comes from that feature. Not resolved here; it needs a real `helm install` to confirm. Flagging rather than guessing.

## Checklist

- [x] I updated navigation when the document structure changed. Both pages keep their existing sidebar placement — `manage-kafka` under Remote WAL, `deploy-kafka` reachable from the Kubernetes overview and enterprise installation pages — so no navigation change was needed.

### DIFF
--- a/docs/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md
+++ b/docs/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md
@@ -83,9 +83,20 @@ extraConfig: |
   log.flush.interval.ms=1000
   log.retention.hours=4
   log.roll.hours=3
-  log.retention.bytes=250000000
   log.segment.bytes=1073741824
 ```
+
+:::warning Retention settings when this cluster backs Remote WAL
+`log.retention.hours=4` is a testing value. Kafka deletes a segment as soon as
+any retention condition matches, so a retention window shorter than your longest
+flush, failure, or manual-recovery window can remove WAL that GreptimeDB has not
+replayed yet.
+
+Do not add a size-based retention such as `log.retention.bytes` to topics used
+for Remote WAL. See [Required Settings and
+Limitations](/user-guide/deployments-administration/wal/remote-wal/configuration.md#required-settings-and-limitations)
+for the full set of constraints.
+:::
 
 ## Installing Kafka Cluster
 
@@ -166,7 +177,7 @@ kafka-broker-1       1/1     Running   0          8m2s
 kafka-broker-2       1/1     Running   0          8m1s
 kafka-controller-0   1/1     Running   0          8m3s
 kafka-controller-1   1/1     Running   0          8m2s
-kafka-controller-0   1/1     Running   0          8m1s
+kafka-controller-2   1/1     Running   0          8m1s
 ```
 </details>
 
--- a/docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
+++ b/docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
@@ -1,103 +1,26 @@
 ---
 keywords: [Kafka, kubernetes, helm, GreptimeDB, remote WAL, installation, configuration, management]
-description: This guide describes how to install and manage Kafka cluster.
+description: How a Kafka cluster is used as GreptimeDB Remote WAL storage, and where to find its deployment and configuration requirements.
 ---
 # Manage Kafka
 
-The GreptimeDB cluster uses Kafka as the [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) storage. This guide describes how to manage Kafka cluster. This guide will use Bitnami's Kafka Helm [chart](https://github.com/bitnami/charts/tree/main/bitnami/kafka) as an example.
+When [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) is enabled, the GreptimeDB cluster writes its write-ahead log to Kafka instead of local disks. Datanodes append WAL entries to Kafka topics, and Metasrv prunes entries that every region has already flushed to object storage.
 
-## Prerequisites
+This means the Kafka cluster holds data that has not been persisted anywhere else yet. Treat it as a stateful dependency of the database, not as a transport buffer.
 
-- [Kubernetes](https://kubernetes.io/docs/setup/) >= v1.23
-- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) >= v1.18.0
-- [Helm](https://helm.sh/docs/intro/install/) >= v3.0.0
+## Deploy the Kafka cluster
 
-## Install
+Follow [Deploying Kafka Cluster](/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md) for installation, verification, monitoring, and uninstallation with the Bitnami Kafka Helm chart.
 
-Save the following configuration as a file `kafka.yaml`:
+## Requirements for Remote WAL
 
-```yaml
-global:
-  security:
-    allowInsecureImages: true
+Before pointing GreptimeDB at the cluster, review [Required Settings and Limitations](/user-guide/deployments-administration/wal/remote-wal/configuration.md#required-settings-and-limitations). It covers the retention policy, the Kafka permissions Datanode needs, and the relationship between `max_batch_bytes` and Kafka's maximum message size.
 
-image:
-  registry: docker.io
-  repository: greptime/kafka
-  tag: 3.9.0-debian-12-r12
+Getting the retention policy wrong is the most common way to lose data here: Kafka deletes segments on its own schedule, and a WAL entry deleted before GreptimeDB replays it cannot be recovered.
 
-controller:
-  replicaCount: 3
+## Connect GreptimeDB to Kafka
 
-  resources:
-    requests:
-      cpu: 2
-      memory: 2Gi
-    limits:
-      cpu: 2
-      memory: 2Gi
+To configure the GreptimeDB cluster itself, see:
 
-  persistence:
-    enabled: true
-    size: 200Gi
-
-broker:
-  replicaCount: 3
-
-  resources:
-    requests:
-      cpu: 2
-      memory: 2Gi
-    limits:
-      cpu: 2
-      memory: 2Gi
-
-  persistence:
-    enabled: true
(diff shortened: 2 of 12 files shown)


docs/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md
```diff
@@ -83,10 +83,21 @@ extraConfig: |
   log.flush.interval.ms=1000
   log.retention.hours=4
   log.roll.hours=3
-  log.retention.bytes=250000000
   log.segment.bytes=1073741824
 ```
 
+:::warning Retention settings when this cluster backs Remote WAL
+`log.retention.hours=4` is a testing value. Kafka deletes a segment as soon as
```
WenyXu, line 90: `log.retention.hours=4` is still part of the copy-pasteable base configuration, even though this page is now the canonical Kafka deployment guide linked from the Remote WAL docs. Labeling it as a testing value does not prevent it from being applied in production, where Kafka may delete WAL that is still needed for recovery. Could we move this setting to an explicitly test-only example, or remove it from the base values?

docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
```diff
@@ -1,103 +1,26 @@
 ---
 keywords: [Kafka, kubernetes, helm, GreptimeDB, remote WAL, installation, configuration, management]
-description: This guide describes how to install and manage Kafka cluster.
+description: How a Kafka cluster is used as GreptimeDB Remote WAL storage, and where to find its deployment and configuration requirements.
 ---
 # Manage Kafka
 
-The GreptimeDB cluster uses Kafka as the [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) storage. This guide describes how to manage Kafka cluster. This guide will use Bitnami's Kafka Helm [chart](https://github.com/bitnami/charts/tree/main/bitnami/kafka) as an example.
+When [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) is enabled, the GreptimeDB cluster writes its write-ahead log to Kafka instead of local disks. Datanodes append WAL entries to Kafka topics, and Metasrv prunes entries that every region has already flushed to object storage.
```
WenyXu: `prunes entries that every region has already flushed` is not quite the pruning invariant. Metasrv derives a per-topic boundary from each Region's prunable entry ID, and an inactive Region can advance that boundary without flushing those offsets. Could we keep this user-facing and say that Metasrv prunes WAL entries that are no longer needed for recovery?

The next paragraph could likewise say that Kafka may hold the only durable copy of writes that have not yet been flushed to object storage.

docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
```diff
@@ -1,103 +1,26 @@
...
-- [Kubernetes](https://kubernetes.io/docs/setup/) >= v1.23
-- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) >= v1.18.0
-- [Helm](https://helm.sh/docs/intro/install/) >= v3.0.0
+## Deploy the Kafka cluster
 
-## Install
+Follow [Deploying Kafka Cluster](/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md) for installation, verification, monitoring, and uninstallation with the Bitnami Kafka Helm chart.
 
-Save the following configuration as a file `kafka.yaml`:
+## Requirements for Remote WAL
```
WenyXu: This page links to the configuration page three times in 26 lines and splits a short handoff page into several one-paragraph sections. Now that the installation and management steps have moved out, `Manage Kafka` also no longer matches the content. Could we reduce this to one short explanation plus the deployment and configuration links, and rename it to `Kafka for Remote WAL`?

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
```diff
@@ -1,110 +1,27 @@
 ---
 keywords: [Kafka, kubernetes, helm, GreptimeDB, remote WAL, 安装, 配置, 管理]
-description: 本指南介绍如何安装和管理 Kafka 集群。
+description: 介绍 Kafka 集群在 GreptimeDB Remote WAL 中的作用，以及部署和配置要求的查阅位置。
 ---
 
 # 管理 Kafka
 
-GreptimeDB 集群在启用 [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) 时，会使用 Kafka 作为 WAL 存储。本文介绍如何部署和管理 Kafka 集群，使用的是 Bitnami 提供的 Kafka Helm [chart](https://github.com/bitnami/charts/tree/main/bitnami/kafka)。
+启用 [Remote WAL](/user-guide/deployments-administration/wal/remote-wal/configuration.md) 后，GreptimeDB 集群不再把预写日志写到本地磁盘，而是写入 Kafka。Datanode 向 Kafka topic 追加 WAL entry，Metasrv 负责清理所有 region 都已 flush 到对象存储的部分。
```
WenyXu: `WAL entry` and `segment` are not the terms used in the Chinese user docs. `WAL 条目` and `Kafka 日志段` would be more natural here. This paragraph already mixes `topic`, `region`, and `flush`, so translating these two terms would also make it easier to read.

docs/user-guide/deployments-administration/deploy-on-kubernetes/deploy-kafka.md
killme2008: The warning is enough.

docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
killme2008: I think it's not in this PR range.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
killme2008: Better to keep the English term here.

daviderli614 (approved):

docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
killme2008: You're right, and the source confirms it precisely.

`LeaderRegionManifestInfo::prunable_entry_id()` returns `max(flushed_entry_id, topic_latest_entry_id)` (`src/common/meta/src/region_registry.rs:114-120`), so an idle region can advance its own boundary via `topic_latest_entry_id` without ever having flushed those offsets. `find_pruneable_entry_id_for_topic` then takes the minimum across the topic's regions (`src/meta-srv/src/procedure/wal_prune/utils.rs:91-96`). My sentence was wrong on both counts — neither "flushed" nor "every region".

Fixed in e93fcfa4 using your phrasing, and the next paragraph now says Kafka may hold the only durable copy of writes not yet flushed to object storage.

docs/user-guide/deployments-administration/wal/remote-wal/manage-kafka.md
killme2008: Reversing my earlier "not in this PR range" — this whole file is new in this PR, so every line you're pointing at was introduced here. Leaving it would just hand the problem to the next PR.

All three points addressed in b8826ebb:

- **Repeated links**: down to two references to the configuration page, and they now have distinct jobs — the page itself, and its required-settings section nested in the same bullet. The first mention of Remote WAL was linking to a page titled "Configuration", which was the wrong target for a concept; it now points at the WAL overview.
- **One-paragraph sections**: all three H2s removed. The retention warning is folded into the lead paragraph and the destinations are a plain list. 26 lines down to 13.
- **Title**: renamed to `Kafka` rather than `Kafka for Remote WAL`. The sidebar already nests this under a `Remote WAL` category, so the longer name would read as "Remote WAL > Kafka for Remote WAL". Filename is unchanged, so the URL is stable.

WenyXu (approved):

PR #2764, opened 2026-08-18
docs: correct behavior and safety boundaries against the implementation

killme2008:
## What changed

Each item was checked against the GreptimeDB source rather than inferred from a neighbouring page.

| Page | Correction | Evidence |
| --- | --- | --- |
| `disaster-recovery/back-up-&-restore-meta-data.md` | On a non-empty target without `--force`, the command logs a warning and returns **success without restoring anything** — a silent no-op the page did not mention. `--force` only skips that check; it never clears the target, so snapshot keys are merged into whatever is already there. | `src/cli/src/metadata/snapshot.rs:179-183`, `src/common/meta/src/snapshot.rs:180` (batch put only) |
| `maintenance/sequence-management.md` | Recovery mode does **not** prevent table creation. It only gates `set-next-id` and relaxes Datanode startup checks. The procedure is now: block new procedure submissions → wait for in-flight procedures → enable recovery mode → set the ID → restart and re-verify → disable and resume. | `recovery_mode` is read only in `admin/sequencer.rs`, `admin/recovery.rs` and `datanode.rs`; `/admin/procedure-manager/pause` is at `meta-srv/src/service/admin.rs:287` |
| same | Added an executable convergence check against `INFORMATION_SCHEMA.PROCEDURE_INFO`. Pausing rejects only new **root** submissions — the switch is checked solely in `ProcedureManager::submit`, and the runner submits child procedures through a path that never consults it. | `common/procedure/src/local.rs:897-903, 961-967`; `local/runner.rs:486-508`. Status strings come from `ProcedureStatus::as_str_name()` (greptime-proto `meta/procedure.proto:15-23`) |
| `integrations/mcp.md` | The read-only default covers `execute_sql` only. `create_pipeline`, `delete_pipeline`, `create_dashboard` and `delete_dashboard` go through the HTTP API and write regardless of `--allow-write`. | `greptimedb_mcp_server` 0.5.2 `server.py:569` vs `1063/1189/1264/1302` |
| `migrate-to-greptimedb/migrate-from-influxdb.md` | The import loop used `curl` without `--fail`, so a 4xx/5xx left the exit code at 0 and the loop moved to the next file — silent batch loss mid-migration. Now fail-fast. | See verification below |
| `query-data/log-query.md` | 1000 is the default applied when `limit.fetch` is absent, not a maximum. | `src/servers/src/http/logs.rs:28-32` uses `get_or_insert` |
| `performance-tuning/design-table.md` | The skipping index is not equality-only; it also serves `IN` and `OR`-equality chains. | `mito2/src/sst/index/bloom_filter/applier/builder.rs:133,194,230` |
| `reference/sql/create.md` | `CREATE FLOW` grammar was missing `EVAL INTERVAL`, which the example on the same page already used. | Same page |
| `traces/data-model.md` | `parent_span_id` does get a skipping index; the sample DDL omitted it while the prose below claimed otherwise. Also documented that `trace_table_partitions` must be a power of two in `[2, 65536]`. | `src/operator/src/insert.rs:781`, `src/sql/src/partition.rs:62` |
| `deployments-administration/configuration.md` | The GCS provider is `Gcs`; the options table said `Gsc` while the example further down already used `Gcs`. Prose now uses `GCS` and the literal value stays `Gcs`. | `src/object-store/src/config.rs:55` |
| `traces/read-write.md` | `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is used verbatim, and a Collector's path is `/v1/traces`. `/v1/otlp/v1/traces` is GreptimeDB's own path, so sending there would 404. The Collector's exporter still points at `http://greptimedb:4000/v1/otlp`. | OTLP exporter spec |
| `wal/remote-wal/configuration.md`, `deploy-on-kubernetes/configure-remote-wal.md` | `replication_factor = 1` was presented unconditionally. Marked as a testing value and noted it must not exceed the broker count. | — |

## Scope

- Languages: English, Chinese

## Backport

Applied to 1.2 and 1.1. Every runtime claim was re-verified against `release/v1.1` and `release/v1.2` instead of being carried over from `main`: the restore no-op, the `recovery_mode` consumers, the pause check inside `submit`, the `ProcedureStatus` values, `DEFAULT_FETCH`, the bloom filter's `InList` / OR-equality collection, `EVAL INTERVAL` parsing, the `parent_span_id` index, the partition constraint and the `Gcs` provider name all hold in both releases.

The backport patches hunks rather than overwriting files, so 1.1 keeps its own `create.md` table options and `remote-wal/configuration.md` option set, and 1.2 keeps its own event type list.

## Verification

```
git diff --check
DOC_LANG=en pnpm check:links
DOC_LANG=zh pnpm check:links
```

Both link checks pass, across Nightly, 1.2, 1.1 and 1.0.

The fail-fast loop was run against a local server returning 200, 400 and 500:

- current form, 500: loop continued through every file, exit code **0**
- new form, 200: all files uploaded, exit 0
- new form, 400: aborted on the first file, exit 1
- new form, 500: retried per `--retry 3`, then aborted, exit 1

## Review follow-ups

- Dropped the requirement to resolve `Failed` / `Poisoned` before changing the sequence. Those runners have already exited and released their locks (`local/runner.rs:284-292`, then `guard.finish()` at `:181-193`); their rows only linger until background cleanup, so requiring a resolution step could block an otherwise quiescent cluster.
- Retitled the pause step to state what it does **not** do, since the switch never suspends accepted procedures or their children.
- Revised the Chinese added by this PR to read as native technical prose.

## Checklist

- [x] I updated navigation when the document structure changed. No structural change.


docs/user-guide/deployments-administration/maintenance/sequence-management.md
```diff
@@ -34,10 +34,24 @@ You can get or set the `next table ID` using Metasrv's HTTP interface at the fol
 
 To safely update the `next table ID`, follow this step-by-step process:
 
-1. **Enable cluster recovery mode** - This prevents new table creation during the update process. See [Cluster Recovery Mode](/user-guide/deployments-administration/maintenance/recovery-mode.md) for more details.
-2. **Set the next table ID** - Use the HTTP interface to set the `next table ID`.
-3. **Restart metasrv nodes** - This ensures the new `next table ID` is properly applied.
-4. **Disable cluster recovery mode** - Resume normal cluster operations.
+1. **Pause the Procedure Manager** - This rejects new metadata-changing operations, including table creation. See [Prevent Metadata Changes](/user-guide/deployments-administration/maintenance/prevent-metadata-changes.md).
```
killme2008: @WenyXu Please take a look of this processing.

docs/user-guide/deployments-administration/maintenance/sequence-management.md
```diff
@@ -34,10 +34,24 @@ You can get or set the `next table ID` using Metasrv's HTTP interface at the fol
...
+2. **Wait for in-flight procedures to finish** - Pausing only rejects *new* procedures; the ones already running continue. Poll until no procedure is still in progress:
```
evenyag: @WenyXu PTAL

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md
```diff
@@ -391,12 +391,12 @@ GreptimeDB 支持将数据保存在本地文件系统，AWS S3 以及其兼容
...
-|         | bucket            | 字符串 | Gsc 桶名称                                          |
-|         | scope             | 字符串 | Gsc 权限                                            |
-|         | credential_path   | 字符串 | Gsc 访问证书                                        |
+| Gcs     |                   |        | Google Cloud Storage 存储选项，当 type="Gcs" 时有效 |
+|         | name            | 字符串 |  存储提供商名字，默认为 `Gcs`               |
+|         | root              | 字符串 | Gcs 桶中的根路径                                    |
+|         | bucket            | 字符串 | Gcs 桶名称                                          |
+|         | scope             | 字符串 | Gcs 权限                                            |
+|         | credential_path   | 字符串 | Gcs 访问证书                                        |
 |         | endpoint          | 字符串 | GSC 的 API 端点                                     |
```
evenyag:
```suggestion
|         | endpoint          | 字符串 | Gcs 的 API 端点                                     |
```

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/integrations/mcp.md
```diff
@@ -9,7 +9,7 @@ description: 了解如何将 GreptimeDB 与模型上下文协议（MCP）集成
 GreptimeDB MCP Server 目前处于实验阶段并在积极开发中。API 和功能可能会在没有通知的情况下发生变化。请在生产环境中谨慎使用。
 :::
 
-[GreptimeDB MCP Server](https://github.com/GreptimeTeam/greptimedb-mcp-server) 提供了模型上下文协议的实现，使 Claude 等 AI 助手能够安全地用 SQL、TQL（兼容 PromQL）和 RANGE 查询来查询与分析您的 GreptimeDB 数据库——并内置只读保护和数据脱敏。
+[GreptimeDB MCP Server](https://github.com/GreptimeTeam/greptimedb-mcp-server) 提供了模型上下文协议的实现，使 Claude 等 AI 助手能够安全地用 SQL、TQL（兼容 PromQL）和 RANGE 查询来查询与分析您的 GreptimeDB 数据库——并内置只读 SQL 闸门和数据脱敏。
```
evenyag: Is "闸门" a proper translation?

docs/user-guide/deployments-administration/maintenance/sequence-management.md
```diff
@@ -34,10 +34,24 @@ You can get or set the `next table ID` using Metasrv's HTTP interface at the fol
...
+
+   ```sql
+   SELECT procedure_id, procedure_type, status
+   FROM INFORMATION_SCHEMA.PROCEDURE_INFO
+   WHERE status IN ('Running', 'Retrying', 'PrepareRollback', 'RollingBack');
+   ```
+
+   Proceed only when this returns no rows. If procedures keep running past the window you allotted, stop here and investigate instead of changing the sequence. A `Failed` or `Poisoned` row means a procedure stopped in an unclean state; resolve it before continuing. See [PROCEDURE_INFO](/reference/sql/information-schema/procedure-info.md).
```
WenyXu: `Failed` and `Poisoned` are terminal states. The runner has exited and released its locks, but these rows remain in `PROCEDURE_INFO` until background cleanup. There is no corresponding resolution step here, so this requirement can block an otherwise quiescent cluster. Could we remove this prerequisite and use only the four active states in the query as the convergence condition?

docs/user-guide/deployments-administration/maintenance/sequence-management.md
```diff
@@ -34,10 +34,24 @@ You can get or set the `next table ID` using Metasrv's HTTP interface at the fol
...
+3. **Enable cluster recovery mode** - Setting the `next table ID` is only allowed in recovery mode. See [Cluster Recovery Mode](/user-guide/deployments-administration/maintenance/recovery-mode.md) for more details.
+4. **Set the next table ID** - Use the HTTP interface to set the `next table ID`.
+5. **Restart metasrv nodes** - This ensures the new `next table ID` is properly applied. After the restart, confirm that recovery mode is still enabled and that the Procedure Manager is still paused, then verify the new `next table ID`.
```
WenyXu: `Sequence` allocates table IDs in per-process ranges of 10. Updating the shared value resets only the Metasrv instance that handles `set-next-id`; another instance can retain an older local range and reuse it after a leadership change. Could we explicitly require restarting **every Metasrv node** to clear all cached table ID ranges?

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md
WenyXu: `Gcs` is the exact configuration value, while the service acronym is `GCS`. The descriptions in both locale tables currently use `Gcs`, and this row still says `GSC`. Could we keep `Gcs` only for literal values such as `type="Gcs"` and use `GCS` in all prose descriptions?

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-meta-data.md
```diff
@@ -85,7 +85,9 @@ greptime cli meta snapshot save \
 :::warning
 **重要**: 在导入元数据之前，请确保目标存储后端的对应表中没有**任何数据**，否则可能会导致元数据损坏。
 
-如果你需要导入到具有现有数据的后端，请使用 `--force` 标志绕过此安全检查。但是，请谨慎操作，因为这可能导致数据损坏。
+如果目标后端非空且**没有**传 `--force`，命令会打印一条警告并直接退出，不会恢复任何数据。此时退出码仍然是成功，所以必须查看命令输出，不能默认恢复已经执行。
+
+`--force` 只是绕过非空检查，并不会先清空目标：恢复过程只写入 snapshot 中包含的 key，目标中其余已有的 key 会原样保留——这正是恢复后的集群会混入两代元数据的原因。建议恢复到空的 backend，之后重启整个集群并验证 catalog、table route 和 procedure。
```
WenyXu: Could we use the established Chinese terms here instead of mixing several English terms into one sentence? “混入两代元数据” is also less direct than describing the possible metadata inconsistency.

```suggestion
`--force` 只会跳过非空检查，不会清空目标后端。恢复时，快照中的键会直接写入，其他已有键仍会保留，因此可能导致元数据不一致。建议恢复到空后端，并在重启集群后检查 catalog、表路由和 Procedure 状态。
```

docs/user-guide/traces/read-write.md
```diff
@@ -86,7 +86,7 @@ OpenTelemetry Collector. For example, you can use the environment variable
 `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` to configure the endpoint of the exporter:
 
 ```shell
-export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://localhost:4318/v1/otlp/v1/traces"
+export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://localhost:4318/v1/traces"
```
evenyag, line 89: @shuiyisong PTAL

I found that both `v1/otlp/v1/traces` and `v1/traces` exist in our code. What's the difference?

killme2008: This is the configuration of otel-collector, it's not ours.

docs/user-guide/deployments-administration/maintenance/sequence-management.md
WenyXu: “Pause the Procedure Manager” can sound as though it suspends procedures that are already running. The pause switch only rejects new root procedure submissions; accepted procedures and their child procedures continue running. Could we rename this step to “Block new procedure submissions” and explicitly state that already-accepted procedures are not suspended or cancelled?

docs/user-guide/deployments-administration/maintenance/sequence-management.md
killme2008: Agreed, and the source backs your reading. `execute_procedure_in_loop` returns on `Done`, `Failed` and `Poisoned` alike (`common/procedure/src/local/runner.rs:284-292`), and the caller then runs `guard.finish()` to release the locks — the comment right above it says the lock must be released "whenever the execution is successful or fail" (`:181-193`). The metadata is deliberately kept afterwards so callers can still query the state, which is exactly why those rows linger.

So requiring them to be resolved could stall a cluster that is already quiescent, with no resolution step to point at. Removed in ddf96a81; the paragraph now explains why terminal states are excluded rather than asking for action on them.

For the record, a parallel review suggested the opposite — adding `Failed` and `Poisoned` *into* the filter. That would be worse: the query would never return empty while those rows are awaiting cleanup.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md
killme2008: Applied in ddf96a81. `Gcs` is now kept only for literal values — the provider row, `type="Gcs"` and the `default is \`Gcs\`` cell — and every prose description in both locales uses `GCS`.

The uppercase `GSC` survived my first pass because I replaced with a word-boundary match on `Gsc`, which does not match `GSC`. Thanks for catching it.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-meta-data.md
killme2008: Taken as suggested in ddf96a81. You are right that the original mixed too many English terms into one sentence, and "混入两代元数据" was an indirect way of saying the metadata can end up inconsistent.

i18n/zh/docusaurus-plugin-content-docs/current/user-guide/integrations/mcp.md
killme2008: It is not — replaced with 校验 in ddf96a81.

To be clear about the scope: 闸门 was already on this page at the "安全闸门" bullet and in the write-mode paragraph; I only followed it in the intro line. Rather than leave two terms for one concept, I changed all three occurrences, so that commit touches two lines this PR had not otherwise modified.

docs/user-guide/traces/read-write.md
killme2008: Both paths do exist, but they belong to different servers. `/v1/otlp/v1/traces` is GreptimeDB's own OTLP endpoint. The line under review sets `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` for an exporter pointing at the **OpenTelemetry Collector** started earlier in the same section, and a Collector's OTLP/HTTP receiver serves `/v1/traces`.

`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is the signal-specific variable, so the URL is used verbatim rather than having a signal path appended. Sending to `/v1/otlp/v1/traces` on the Collector would 404.

The GreptimeDB endpoint on the same page is unchanged — the Collector's own exporter still points at `http://greptimedb:4000/v1/otlp`.

docs/user-guide/deployments-administration/maintenance/sequence-management.md
killme2008: Correct on both counts, and the split is visible in the code.

The pause switch is consulted in exactly one place — `check_status()` at `common/procedure/src/local.rs:897-903`, called only from `ProcedureManager::submit` (`:961-967`). Child procedures never reach it: the runner submits them through its own path at `local/runner.rs:486-508`, which checks only whether the manager is stopped and whether the procedure was already accepted.

Retitled the step to "Block new procedure submissions" and moved the caveat into it, so the guarantee and its limit are stated together instead of the limit appearing one step later. Pushed as 25658cdf.

WenyXu (approved):

evenyag (approved):

PR #2765, opened 2026-08-18
docs: correct deployment resource names and integration prerequisites

killme2008:
### Deployment

| Page | Correction | Evidence |
| --- | --- | --- |
| `monitoring/cluster-monitoring-deployment.md` | The page used `${cluster-name}-monitoring` for the CR and `${cluster-name}-monitoring-standalone` for the Service, while other lines on the same page already used `-monitor-standalone`. The operator creates a `GreptimeDBStandalone` named **`${cluster}-monitor`**, and its Service/StatefulSet/Pods are named **`${cluster}-monitor-standalone`**. These are two different names and the page now says so. | operator `controllers/common/common.go:306-308` (`MonitoringServiceName` returns `name + "-monitor"`), used as the CR object key at `greptimedbcluster/deployers/monitoring.go:90-93`; the workload DNS at `:123` wraps it in `ResourceName(..., StandaloneRoleKind)` |
| same | The PVC cleanup `danger` box said the operation removes the GreptimeDB cluster's metadata and data. It deletes the monitoring standalone's PVCs. The command also hardcoded `-n default` while using `${cluster-name}`. | Same page |
| `common-helm-chart-configurations.md` | The S3 section was titled "AWS EKS Pod Identity" and glossed it as "IAM Roles for Service Accounts", but the example annotates a service account with `eks.amazonaws.com/role-arn`, which is IRSA. Pod Identity uses a pod identity association and no such annotation. Retitled to IRSA and noted the distinction; the published slug is preserved with `AnchorAlias`. | AWS docs for both mechanisms |
| `deploy-minio.md` | Install output was missing the port ("accessed via port  on"); the prerequisite said PodMonitor CRD while the example configures `serviceMonitor`; the sample root password was presented as fixed login credentials. | Same page |

### Integrations

| Page | Correction | Evidence |
| --- | --- | --- |
| `mindsdb.md` | "work out of box" no longer holds. The GreptimeDB handler moved to `mindsdb/community-handlers`, and community handlers are disabled by default — `MINDSDB_COMMUNITY_HANDLERS=true` plus a restart is required. The `ENGINE = 'greptimedb'` syntax itself is unchanged, so the example stays. | `mindsdb/mindshub@490c5f0` removed `greptimedb_handler/*`; the handler and its `ENGINE` syntax live in `mindsdb/community-handlers`; the env var is documented in that repo's README |
| `perses.md` | Dropped the claim that panels and variables "should work without changes". Upstream states the migration is best-effort, supports Grafana 9.0–11.x, does not migrate alerts or users, and replaces unsupported panels/variables with static placeholders. | perses.dev migration docs |
| `superset.md` | Marked Docker Compose as evaluation-only (upstream: "we do not support nor recommend using our `docker compose` constructs to support production-type use-cases") and documented the SQLAlchemy constraint. | `greptimedb-sqlalchemy` 0.1.7 requires `sqlalchemy<2,>=1.4`; Superset 6.1.0 ships 1.4.54 (**compatible today**), master has moved to 2.0.51 |
| `grafana.md` | `Duration Unit` default was `nano_seconds`, which is not a valid value. | Plugin `src/types/queryBuilder.ts` `TimeUnit` enum and `DurationUnitSelect.tsx` accept only `seconds`/`milliseconds`/`microseconds`/`nanoseconds` |

## Coverage

All 19 pages under `user-guide/integrations/` were checked against their upstreams this round, not sampled. Results for the ones **not** changed:

- **flink, spark** — connector identifier `greptimedb` and every documented option (`endpoints`, `database`, `table`, `time-index`, `tags`, `batch.max-rows`, `query.jdbc-url`, `query.fetch-size`) match the connector sources. `io.greptime:{flink,spark}-connector-greptimedb:0.1.0` and its `shaded` classifier are both present on Maven Central.
- **dbeaver, streamlit, metabase** — ports 4002 (MySQL) and 4003 (PostgreSQL) match `config/standalone.example.toml:127,162`; the Metabase driver asset is still named `greptimedb.metabase-driver.jar` in the latest release.
- **coroot** — uses the `?db=<db-name>` query parameter, which matches the canonical form in the Prometheus ingest docs. No header change needed.
- **alloy, emqx, fluent-bit, kafka, prometheus, telegraf, vector, overview** — pure redirect pages with no configuration of their own; their link targets are covered by the link checker.
- **mcp** — corrected separately in #2764.

`integrations/overview.md` still opens with "seamlessly integrated" / "comprehensive guidance". That is a wording issue, not a factual one, and belongs to the deferred wording pass rather than this PR.

## Scope

- Documentation versions: Nightly, 1.2, 1.1 (Perses: Nightly and 1.2 only — the page does not exist in 1.1)
- Languages: English, Chinese

## Verification

```
git diff --check
DOC_LANG=en pnpm check:links
DOC_LANG=zh pnpm check:links
```

Both link checks pass, including the `AnchorAlias` added for the renamed IRSA heading.

## Still open

The Grafana plugin install section documents `info8fcc-greptimedb-datasource.zip`, while the plugin README instructs downloading `info8fcc-greptimedb-datasource-unsigned.zip` and setting `allow_loading_unsigned_plugins`. Both assets exist in the latest release and I could not determine from release metadata whether the signed one loads without the allowlist. Left unchanged pending confirmation from the plugin maintainer.

## Checklist

- [x] I updated navigation when the document structure changed. No structural change.

### DIFF
--- a/docs/user-guide/integrations/superset.md
+++ b/docs/user-guide/integrations/superset.md
@@ -14,7 +14,9 @@ follow this guide.
 ### Running Superset with Docker Compose
 
 [Docker compose](https://superset.apache.org/docs/installation/docker-compose)
-is the recommended way to run Superset. To add GreptimeDB extension, create a
+is the quickest way to try Superset locally. Upstream states it does not support
+or recommend the `docker compose` constructs for production use, so treat this
+path as evaluation only. To add GreptimeDB extension, create a
 `requirements-local.txt` file in `docker/` of Superset codebase.
 
 Add GreptimeDB dependency in `requirements-local.txt`:
@@ -38,6 +40,13 @@ to the same environment.
 ```bash
 pip install greptimedb-sqlalchemy
 ```
+
+:::note SQLAlchemy version
+`greptimedb-sqlalchemy` currently requires SQLAlchemy 1.x. Superset 6.1.0 and
+earlier pin SQLAlchemy 1.4, so the two are compatible today. Superset's main
+branch has moved to SQLAlchemy 2.0, so a later Superset release will need a
+dialect that supports 2.0 first.
+:::
 
 ## Add GreptimeDB as database
 
--- a/docs/user-guide/integrations/superset.md
+++ b/docs/user-guide/integrations/superset.md
@@ -42,10 +42,11 @@ pip install greptimedb-sqlalchemy
 ```
 
 :::note SQLAlchemy version
-`greptimedb-sqlalchemy` currently requires SQLAlchemy 1.x. Superset 6.1.0 and
-earlier pin SQLAlchemy 1.4, so the two are compatible today. Superset's main
-branch has moved to SQLAlchemy 2.0, so a later Superset release will need a
-dialect that supports 2.0 first.
+`greptimedb-sqlalchemy` requires `sqlalchemy>=1.4,<2`. Superset 6.1.0, the
+current release, pins SQLAlchemy 1.4.54 and works with it. Superset's main
+branch has moved to SQLAlchemy 2.0, so a later release will need a dialect that
+supports 2.0 first. If you run an older Superset, check its own SQLAlchemy pin
+against that range.
 :::
 
 ## Add GreptimeDB as database
--- a/docs/user-guide/deployments-administration/deploy-on-kubernetes/common-helm-chart-configurations.md
+++ b/docs/user-guide/deployments-administration/deploy-on-kubernetes/common-helm-chart-configurations.md
@@ -325,9 +325,13 @@ objectStorage:
     endpoint: ""
 ```
 
-#### Using AWS EKS Pod Identity for S3
+<AnchorAlias id="using-aws-eks-pod-identity-for-s3" />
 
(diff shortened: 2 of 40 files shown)


docs/user-guide/integrations/superset.md
```diff
@@ -39,6 +41,13 @@ to the same environment.
 pip install greptimedb-sqlalchemy
 ```
 
+:::note SQLAlchemy version
+`greptimedb-sqlalchemy` currently requires SQLAlchemy 1.x. Superset 6.1.0 and
+earlier pin SQLAlchemy 1.4, so the two are compatible today. Superset's main
+branch has moved to SQLAlchemy 2.0, so a later Superset release will need a
+dialect that supports 2.0 first.
```
killme2008: Do we have a plan to support SQLAlchemy 2.0 in `greptimedb-sqlalchemy`? @sunng87

sunng87: We can check and update in near future.

sunng87 (approved):

daviderli614 (approved):

sunchanglong (approved):

PR #2766, opened 2026-08-18
docs: document two-phase series scan mode

### LINKED ISSUE #2714: Update docs for feat(mito2): introduce two-phase metric series scans
A document change request is generated from https://github.com/GreptimeTeam/greptimedb/pull/8826

evenyag:
## What changed

- Document the `region_engine.mito.experimental_series_scan_v2` option in the English and Chinese Nightly configuration guides.
- Document the `SeriesScan.mode` field and its `two_phase` and `legacy` values in the English and Chinese `EXPLAIN` references.

Closes #2714.

## Scope

- Documentation versions: Nightly
- Languages: English, Chinese

The feature merged after the v1.2 documentation cut, so released documentation is unchanged.

## Verification

- `git diff --check`
- Compiled all four changed Markdown files with `@mdx-js/mdx`
- `pnpm build` and `DOC_LANG=zh pnpm build` were attempted, but the host exhausted its process table and Docusaurus failed while spawning Git history lookups for unrelated pages with `EAGAIN`.

## Checklist

- [x] I checked changed links and anchors; this change adds none.
- [x] I updated navigation when the document structure changed; navigation is unchanged.


docs/reference/sql/explain.md
```diff
@@ -485,4 +486,7 @@ sort operators when a query needs ordered output.
 
 `SeriesScan` returns rows grouped by series. It uses the same per-partition
 metric structure as the other scanners and can also report distributor metrics
-from the series distributor partition.
+from the series distributor partition. GreptimeDB automatically selects an
+eligible scan mode and reports the result in the `mode` field. Setting
+`region_engine.mito.experimental_series_scan_v2` to `false` forces `legacy`
+mode.
```
fengjiachun: This leaves `mode` unpredictable. `experimental_series_scan_v2` defaults to `true`, but `SeriesScan::new` picks `two_phase` only when `supports_two_phase` also holds: the region metadata must use sparse `(__table_id, __tsid)` primary keys — i.e. a metric engine physical region — and, in enterprise builds, the scan must have no extension ranges. Ordinary tables therefore always print `"mode":"legacy"` under the default configuration, which is exactly what the sqlness expectations updated by GreptimeTeam/greptimedb#8826 record: `"distribution":"PerSeries", "mode":"legacy"`.

Read as written — "automatically selects an eligible scan mode" plus "`false` forces `legacy`" — a user will infer that the `true` default yields `two_phase`, and treat the `legacy` they actually see as a misconfiguration. Please state what makes a scan eligible: `two_phase` needs a metric engine physical region *and* the option enabled; every other `SeriesScan` reports `legacy`.

Same text in `i18n/zh/docusaurus-plugin-content-docs/current/reference/sql/explain.md` lines 476-479.

docs/user-guide/deployments-administration/configuration.md
```diff
@@ -651,6 +652,7 @@ Available options:
...
 | `default_flat_format`                | Bool    | `true`       | Whether to enable flat format as the default SST format.                                                                                                                                                                                                                                                                                                             |
+| `experimental_series_scan_v2`       | Bool    | `true`       | Whether to enable the experimental two-phase mode for eligible Metric Engine series scans. Set to `false` to use the legacy mode.                                                                                                                                                                                                                                    |
```
fengjiachun: "eligible" is the load-bearing word in this description and it is never defined anywhere in the docs, so a reader cannot tell why a `true` default still produces `legacy`. The condition is `is_sparse_metric_metadata`: the region is a metric engine physical region (this file already notes below that sparse primary key encoding is always on for the metric engine from v1.2, so in practice that is the whole condition), plus no extension ranges in enterprise builds. Spelling that out here is what makes the option actionable — otherwise flipping it is guesswork.

Separately, this file writes "metric engine" in lower case elsewhere (the `with_metric_engine` row and the sparse-encoding note); "Metric Engine" here is inconsistent with that.

Both points apply to `i18n/zh/docusaurus-plugin-content-docs/current/user-guide/deployments-administration/configuration.md` line 659.

evenyag: Updated both configuration references in c43a944d. The option now explicitly applies to series scans of metric engine physical regions, other scans are documented as using `legacy`, and the metric engine capitalization is consistent. I intentionally left out enterprise-specific implementation details because this is an experimental feature.

docs/reference/sql/explain.md
evenyag: Updated the English and Chinese EXPLAIN docs in c43a944d. They now state that `two_phase` is used only for metric engine physical-region scans when the option is enabled, while other `SeriesScan` nodes report `legacy`. I kept the explanation at this concise user-facing level because the feature is experimental, without adding enterprise-specific internal conditions.

fengjiachun (approved):
LGTM

PR #2767, opened 2026-08-18
docs: replace unconditional guarantees with their actual conditions

killme2008:
## What changed

Four places where the docs stated something unconditionally that only holds under conditions the reader was never told about.

**`query-data/view.md`** — the page listed "granting read permissions while restricting columns and rows" as a use case, with no mention of what makes that work. It holds only with a permission model that can grant a view without its base tables. The static user provider grants per user rather than per object, so a view alone isolates nothing there; Enterprise ACLs match the object names that appear in the query, so granting only the view does confine the user to the columns and rows it selects. The use case now carries that condition.

**`integrations/grafana.md`** — the page said the plugin "can currently only be installed on a local Grafana instance" without explaining why or offering an alternative. The plugin is not in the Grafana catalog, so the unsigned archive is the path users actually take, which is also the only one the upstream README documents. The section now leads with it, and the download link and `grafana cli` command point at the unsigned zip instead of the signed one. A signed build bound to your own `root_url` is available on request.

**`disaster-recovery/overview.md`**, **`faq-and-others/faq.md`** — `RPO=0` and `minute-level RTO` were presented as properties of the topology. They are targets, and they depend on conditions the pages did not state:

- Region Failover is disabled by default (`enable_region_failover: false`).
- Kafka, object storage, the metadata backend and the traffic entry point all have to span the failure domain being protected against.
- A standalone instance is **not** stateless with remote WAL and object storage. It keeps catalogs, schemas and table definitions in a local KV store under `<data_home>/metadata` (`standalone.rs` → `metadata_store_dir`), which Kafka and object storage cannot rebuild. That has to be backed up separately.

**Migration guides** (ClickHouse, MySQL, PostgreSQL, InfluxDB) — dual-write was described as a way to avoid data loss. Two independent writes are not a transaction; one side can fail and leave a gap or a duplicate. The strategy stays, with the cutoff, the failure record, the independent retry and the reconciliation. The cutoff is now applied where it matters:

- The ClickHouse export filters on it, and the cutoff must be the instant dual-write started — anything later re-imports rows dual-write already delivered.
- The MySQL and PostgreSQL commands are labelled as full exports, since they cannot filter by time on their own.
- `influx_inspect export` and `influxd inspect export-lp` both skip only when `ts > end`, so `--end` is inclusive and needs an instant just before the cutoff.

## Scope

- Languages: English, Chinese

## Verification

```
git diff --check
DOC_LANG=en pnpm check:links
DOC_LANG=zh pnpm check:links
```

Both pass.

The Grafana signature claim was checked by unpacking both archives from v3.0.3: the signed one carries a `MANIFEST.txt` with `"signatureType": "private"` and `"rootUrls": ["http://localhost:3000/"]`, which is why it does not load elsewhere.


MichaelScofield (approved):
for dual-write when on greptimedb, I think it's worth noting the reader that append-only table could cause duplication so don't retry without check, in case raising some unnecessary surprise

PR #2768, opened 2026-08-18
docs: tighten prose in user guide pages flagged by the writing audit

killme2008:
## What changed

Writing-quality pass over the pages in the audit's section 9 list that the
previous four PRs did not touch. No technical claim was changed; every hunk
removes marketing phrasing, a restated summary, or a grammar slip.

- **Marketing phrasing replaced with what the tool actually does**:
  `integrations/overview.md` ("seamlessly integrated" / "comprehensive
  guidance" in front of a card list), `deploy-on-kubernetes/overview.md`
  ("from day one", "robust metadata storage capabilities for enhanced
  reliability and performance"), `flow-computation/continuous-aggregation.md`
  ("empowers developers", "seamlessly", "invaluable tool", TQL "powerful
  analytical functions for real-time insights"), `deployments-administration/overview.md`
  ("ensure business continuity", "for optimal efficiency").
- **Vendor taglines copied from upstream marketing pages** rewritten as
  factual one-liners: `coroot.md` ("supercharged with predefined dashboards"),
  `streamlit.md` ("a faster way to build and share data apps"), `dbeaver.md`
  ("a popular choice among developers and database administrators").
- **Paragraphs that restate the preceding paragraph** removed:
  `continuous-aggregation.md` (the "In all these usecases..." summary after
  the three use cases, plus the "Consequently, ... real-time insights"
  closer), `flow-computation/overview.md` ("a clever materialized views ...
  with minimal effort"), `manage-data/overview.md`, `deploy-on-kubernetes/overview.md`.
- **Empty section** `## Overview` in `back-up-&-restore-data.md` removed. No
  page links to that anchor.
- **Grammar**: `metabase.md` ("is an open source BI tool that written in
  Clojure", "use you own settings"), `fluent-bit.md` ("as a Output"),
  `use-custom-pipelines.md` ("Once understand the log format"),
  `continuous-aggregation.md` (the run-on "So you can be made aware ..."
  sentence).

Chinese pages carry the same changes where the Chinese text had the same
problem. Some did not — `fluent-bit.md` and `metabase.md` were already
correct in Chinese, so they are English-only.

Section headings, code, SQL, and command output are untouched.

Two factual problems found while reading are deliberately **not** in this PR
and will be filed separately:

- `flow-computation/overview.md` names the source table `nginx_access_log`
  while the `CREATE TABLE` and all later queries use `ngx_http_log`.
- `deployments-administration/overview.md` links "Metadata Storage" to
[...]

PR #2770, opened 2026-08-18
docs: fix inaccurate table names, links, and claims in the user guide

killme2008:
## What changed

Five factual problems found while doing the prose pass in #2768, kept separate
so the wording diff there stayed reviewable.

1. **`flow-computation/overview.md` named the wrong source table.** The prose
   said `nginx_access_log` in two places, while the `CREATE TABLE`, the
   `CREATE FLOW`, the `INSERT`, and every reference after it use
   `ngx_http_log`. Following the page as written creates a flow over a table
   that does not exist. The Chinese page had the same mistake in four places.

2. **`deployments-administration/overview.md` linked "Metadata Storage" to the
   wrong page.** It pointed at `./manage-data/overview.md`, which is data
   management and is already the target of the "Data Management" bullet three
   lines below. Now points at `./manage-metadata/overview.md`. The link checker
   passed before this fix because the old target exists — it was the wrong
   page, not a dead one.

3. **`use-custom-pipelines.md` described a `vrl` behavior the example does not
   have.** The text said the processor sets table options "such as adding a
   suffix to the table name and setting the TTL", but the example only sets
   `.greptime_ttl`. The sentence now describes what this example does.
   `greptime_table_suffix` and the rest of the options stay documented in
   [Write Log API](/reference/pipeline/write-log-api.md), which the page
   already links at the end.

4. **`back-up-&-restore-data.md` reused export headings under Import
   Operations.** `Full Databases Backup`, `Specific Database Backup`, and
   `Schema-Only Operations` each appeared twice, so the import ones published
   as `#full-databases-backup-1`, `#specific-database-backup-1`, and
   `#schema-only-operations-1`. The import headings are renamed to
   `Full Databases Import`, `Specific Database Import`, and
   `Schema-Only Import`, with `AnchorAlias` preserving all three published
   fragments. The Chinese page had the first two (`完整数据库备份-1`,
   `指定数据库备份-1`) and already used distinct headings for the schema-only
   pair. Nothing in the repository links to these fragments, but they are
   published URLs.

5. **`deploy-on-kubernetes/overview.md` stated Remote WAL as a guarantee.**
   "persistently record every data modification and ensure no loss of
   memory-cached data" is the same class of claim #2767 addresses. Now states
   the mechanism: writes a Datanode has not yet flushed can be replayed from
   Kafka after a failure. The Kafka retention conditions that govern whether
   replay is possible are covered on the Remote WAL pages by #2762.

## Scope

- Languages: English, Chinese

## Verification

- Rebased onto `main` after #2768 merged; no conflicts.
- Compared the nightly hunks against the 1.1 and 1.2 copies for every changed
  file: identical, so no hunk context overwrote version-specific text.
- `DOC_LANG=en pnpm check:links` — pass
- `DOC_LANG=zh pnpm check:links` — pass
- Inspected the built HTML for the renamed page. English renders
  `full-databases-backup-1`, `specific-database-backup-1` and
  `schema-only-operations-1` alongside the new import anchors; Chinese renders
  `完整数据库备份-1` and `指定数据库备份-1` alongside `完整数据库导入` and
  `指定数据库导入`. All published fragments still resolve.
- `git diff --check` — clean

## Checklist

### DIFF
--- a/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
+++ b/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
@@ -86,8 +86,10 @@ greptime cli data export \
 
 ## Import Operations
 
-### Full Databases Backup
-Import all databases backup.
+<AnchorAlias id="full-databases-backup-1" />
+
+### Full Databases Import
+Import every database from a backup directory produced by `greptime cli data export`.
 ```bash
 # Import all databases
 greptime cli data import \
@@ -114,7 +116,9 @@ greptime cli data import \
     --target schema
 ```
 
-### Specific Database Backup
+<AnchorAlias id="specific-database-backup-1" />
+
+### Specific Database Import
 ```bash
 # The same applies to import tool
 greptime cli data import \
--- a/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
+++ b/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
@@ -122,7 +122,7 @@ greptime cli data import \
 
 ### Specific Database Import
 ```bash
-# The same applies to import tool
+# Import a specific database
 greptime cli data import \
     --addr localhost:4000 \
     --input-dir /tmp/backup/greptimedb \
--- a/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
+++ b/docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
@@ -106,7 +106,9 @@ greptime cli data import \
     --auth-basic <username>:<password>
 ```
 
-### Schema-Only Operations
+<AnchorAlias id="schema-only-operations-1" />
+
+### Schema-Only Import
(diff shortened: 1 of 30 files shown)


sunng87 (approved):

sunng87:
backport! 1.2

docs/user-guide/deployments-administration/disaster-recovery/back-up-&-restore-data.md
```diff
@@ -114,7 +118,9 @@ greptime cli data import \
     --target schema
 ```
 
-### Specific Database Backup
+<AnchorAlias id="specific-database-backup-1" />
+
+### Specific Database Import
 ```bash
 # The same applies to import tool
```
fengjiachun: Stale comment left over from the rename. `# The same applies to import tool` only made sense while this heading was a duplicate of `### Specific Database Backup` in the export section — it told the reader which example this block mirrored. Now that the heading says `Import`, the back-reference points at nothing.

Every sibling block on this page is self-describing (`# Import all databases`, `# Import only schemas`), and the Chinese page already has `# 导入指定数据库` at this exact line, so EN and ZH now diverge in the one section this PR renamed.

```suggestion
# Import a specific database
```

Same line in `versioned_docs/version-1.1` and `versioned_docs/version-1.2`.

fengjiachun (approved):

PR #2784, opened 2026-08-25
docs: correct outdated Nightly contributor guide content

killme2008:
## What changed

Correct and complete the Nightly contributor guide while keeping it focused on component contracts, main execution paths, and contributor decisions rather than source navigation.

- Replace the early Metasrv design draft and speculative proto examples with current metadata, procedure, leader-election, heartbeat, failover, control-path, and table-route concepts.
- Explain both Flownode execution paths: batching is the primary path for aggregation and TQL workloads; streaming is retained for compatibility and deprecated for new workloads.
- Clarify Frontend request paths, Datanode Region dispatch, Mito WAL recovery boundaries, query planning and execution, partition routing and conservative pruning, and the conditional creation of Puffin index files.
- Turn the testing pages into a test-selection guide covering package tests, Sqlness, integration tests, compatibility tests, and result review.
- Document the current gRPC ingester contract: generated stubs, request representations, success and error metadata, bounded retries, and ambiguous delivery for unary and streaming writes.
- Add the missing Metasrv node-lease, recovery, and table-ID sequence endpoints, including their operational constraints.
- Use the official Apache Parquet file-layout diagram and identify its source.
- Keep the default selector recommendation, intentional overview links, and DeepWiki as a labeled secondary reference.
- Remove the obsolete Python Scripts pages after the implementation was removed.
- Tighten a small number of inflated or awkward sentences without expanding their technical scope.

The architecture and testing boundary follows established contributor documentation practice: explain abstractions, flows, invariants, and how to validate a change; leave implementation-level exploration to the source. No source directory or file-path navigation is added to the architecture pages.

## Scope

- Documentation versions: Nightly, v1.2, v1.1, and v1.0. Each released version was gated on what that
  GreptimeDB release actually ships:
  - v1.2 (v1.2.0-beta.2) takes the change unchanged.
  - v1.1 (v1.1.4) drops the compatibility-test entry (`tests/compatibility` and `cargo sqlness compat`
    landed after that release) and `x-greptime-err-retry-hint` (not sent by v1.1).
  - v1.0 (v1.0.2) additionally omits the new Memtable page, because `MemtableOptions` there still has a
    `PartitionTree` variant, so the page's selection rules do not hold; links to it degrade to plain text.
    It also drops the pending-batching-flow sentence (`defer_on_missing_source` does not exist yet).
  - `reference/sql/create.md` and the Mito memtable configuration section change only in v1.1 and v1.2.
    v1.0.2 still has `MitoConfig::memtable` and accepts `memtable.type=partition_tree`, so the original
    text is correct there.
- Languages: English, Chinese

## Verification

- `typos` on all modified Markdown files
- `git diff --check`
- Verified version-sensitive claims against GreptimeDB main at `0cc83c4570`, the current `greptime-proto`
  definitions, and tags `v1.0.2`, `v1.1.4`, and `v1.2.0-beta.2` for the backported versions.
- Reviewed the complete diff relative to `main`; no released/0.x docs, generated outputs, or lockfiles changed. Two user-guide edits are intentional and in scope: `user-guide/deployments-administration/configuration.md` drops the removed `[region_engine.mito.memtable]` section, and `reference/sql/create.md` corrects the `memtable.type` values.

## Checklist

- [x] I updated navigation when the document structure changed. `sidebars.ts` adds the new Memtable page. The removed Python Scripts pages needed no navigation change because they were already excluded from the build.


docs/contributor-guide/flownode/dataflow.md
```diff
@@ -1,17 +1,38 @@
...
+Flownode has two internal execution paths:
+
+- **Batching mode** is the primary path for aggregation and TQL workloads. It evaluates queries over persisted source data and writes materialized results to a sink table.
+- **Streaming mode** is the legacy path retained for compatibility and deprecated for new workloads. It incrementally processes rows mirrored from Frontend as they arrive.
+
+Users do not select the mode directly. When a Flow is created, GreptimeDB chooses the path from the query and source-table properties. Aggregation, `DISTINCT`, and TQL queries use batching mode. Simple non-aggregation queries, and any Flow whose source table has `ttl = 'instant'`, currently use streaming mode. A Flow deferred because its source table does not yet exist starts as a pending batching Flow.
+
+## Batching mode
+
+Batching mode reuses GreptimeDB's query engine instead of maintaining an operator graph for every incoming row. For a time-windowed Flow, its main loop is:
```
killme2008, line 17: Please take a look at this document that describes the dataflow of Flow engine. @discord9

docs/contributor-guide/metasrv/admin-api.md
```diff
@@ -9,24 +9,25 @@ description: Details the Admin API for Metasrv, including endpoints for health c
...
-The Admin API provides a simple way to view and manage cluster information, including metasrv health detection, metasrv leader query, datanode heartbeat detection, maintenance mode, and Procedure Manager controls.
-
-The Admin API is an HTTP service that provides a set of RESTful APIs that can be called through HTTP requests. The Admin API is simple, user-friendly and safe.
+The Admin API exposes Metasrv status, cluster controls, and metadata recovery operations over HTTP. It does not provide authentication, and some endpoints change cluster behavior or metadata allocation. Deployments must protect the HTTP port with network-level controls.
 This page covers the following APIs:
 
 - /health
 - /leader
 - /heartbeat
+- /node-lease
```
killme2008, line 18: Please take a loot at this document @WenyXu

docs/contributor-guide/metasrv/overview.md
```diff
@@ -1,161 +1,86 @@
...
-![meta](/meta.png)
-
 ## What's in Metasrv
 
-- Store metadata (Catalog, Schema, Table, Region, etc.)
-- Request-Router. It tells the Frontend where to write and read data.
-- Load balancing for Datanode, determines who should handle new table creation requests, more precisely, it makes resource allocation decisions.
-- Election & High Availability, GreptimeDB is designed in a Leader-Follower architecture, only Leader nodes can write while Follower nodes can read, the number of Follower nodes is usually >= 1, and Follower nodes need to be able to switch to Leader quickly when Leader is not available.
-- Statistical data collection (reported via Heartbeats on each node), such as CPU, Load, number of Tables on the node, average/peak data read/write size, etc., can be used as the basis for distributed scheduling.
+Metasrv is the metadata and coordination service in a distributed GreptimeDB cluster. It does not sit on the data path. Its main responsibilities are:
```
killme2008, line 10: Please take a look at this document @WenyXu

docs/contributor-guide/metasrv/overview.md
```diff
@@ -1,161 +1,93 @@
...
-          mutate_endpoint
-          select_endpoint_1, select_endpoint_2
-        region_2
-          mutate_endpoint
-          select_endpoint_1, select_endpoint_2, select_endpoint_3
-        region_xxx
-   table_B
-      ...
+
+A table route maps each Region to its current Datanode peer. It does not contain a separate list of read replicas:
```
WenyXu: Table routes can carry read followers. At `0cc83c4570`, `RegionRoute` has both `leader_peer` and `follower_peers`, and `TableRoute::try_from_raw` populates the latter from `follower_peer_indexes`. The current statement and one-peer diagram are therefore inaccurate even when a deployment has no followers configured. Could we describe one write leader plus zero or more follower/read peers, and update the Chinese page as well?

docs/contributor-guide/metasrv/admin-api.md
```diff
@@ -155,3 +164,39 @@ The response body uses the following format:
...
+Disable recovery mode after the repair is complete. Use [maintenance mode](/user-guide/deployments-administration/maintenance/maintenance-mode.md) instead when the goal is to suspend automatic cluster operations during planned maintenance.
+
+## /sequence/table HTTP endpoints
+
+These endpoints inspect or repair the table ID sequence:
+
+- `GET /admin/sequence/table/next-id`: return the next table ID without allocating it.
+- `POST /admin/sequence/table/set-next-id`: advance the next table ID.
+
+Setting the sequence requires recovery mode. The new value must be greater than the current value; the API cannot move the sequence backwards.
```
WenyXu: The direct `set-next-id` example is incomplete as an operational procedure. The repository’s [sequence-management guide](/user-guide/deployments-administration/maintenance/sequence-management.md) requires pausing the Procedure Manager, waiting for in-flight procedures, enabling recovery mode, restarting all Metasrv nodes after the change, and only then resuming normal operation. Recovery mode alone does not block DDL. Could we link that canonical runbook here, or retain its safety requirements in this page and the Chinese translation?

i18n/zh/docusaurus-plugin-content-docs/current/contributor-guide/metasrv/overview.md
```diff
@@ -1,162 +1,93 @@
...
-- 数据节点的负载均衡，决定谁应该处理新的表创建请求，更准确地说，它做出资源分配决策。
-- 选举与高可用性，GreptimeDB 设计为 Leader-Follower 架构，只有 leader 节点可以写入，而 follower 节点可以读取，follower 节点的数量通常 >= 1，当 leader 不可用时，follower 节点需要能够快速切换为 leader。
-- 统计数据收集（通过每个节点上的心跳报告），如 CPU、负载、节点上的表数量、平均/峰值数据读写大小等，可用作分布式调度的基础。
+Metasrv 是 GreptimeDB 分布式集群中的元数据和协调服务，不参与数据读写链路。它主要负责：
 
-## 前端如何与 Metasrv 交互
+- 存储 Catalog、Schema、Table、Region、路由和节点元数据；
+- 为新 Region 选择 Datanode，并维护表路由；
+- 选举一个 Metasrv leader 负责协调元数据变更；
+- 通过可恢复的 Procedure 执行 DDL、Region 迁移、故障转移和 repartition；
```
WenyXu: The glossary uses `Repartition (重分区)`, while this Chinese sentence leaves lowercase `repartition` in otherwise Chinese prose. Could we use `重分区` or the established `Repartition` spelling to keep the terminology consistent?

docs/contributor-guide/metasrv/admin-api.md
killme2008: Please note, this is contributor guide. If the user want the complete operational doc, they should read references,.

docs/contributor-guide/datanode/memtable.md
```diff
@@ -0,0 +1,93 @@
+---
+keywords: [memtable, Mito engine, write buffer, flush, time partition, BulkMemtable]
+description: How Mito organizes mutable Region data in memtables and moves it into SST files.
+---
+
+# Memtable design
```
killme2008, line 6: This a new document about memtable, please take a look @evenyag

docs/contributor-guide/metasrv/overview.md
killme2008: Fixed in 1be0a8f8. The English and Chinese overviews now describe one leader plus zero or more followers, including the read-replica case, and the route diagram matches that model.

docs/contributor-guide/metasrv/admin-api.md
killme2008: Addressed in 1be0a8f8. The English and Chinese contributor pages keep the API semantics, state that recovery mode is not a DDL barrier, and link the canonical sequence-management runbook for the operational procedure.

i18n/zh/docusaurus-plugin-content-docs/current/contributor-guide/metasrv/overview.md
killme2008: Fixed in 1be0a8f8. The Chinese overview now uses “重分区”.

docs/contributor-guide/flownode/dataflow.md
```diff
@@ -1,17 +1,38 @@
...
+
+1. A source-table write marks the affected time windows as dirty.
+2. A `BatchingTask` wakes on its schedule or after a notification and collects pending dirty windows.
```
discord9: `BatchingTask` is not awakened by dirty-window notifications in the current implementation. The dirty handlers only update `dirty_time_windows`; scheduled tasks wait until the next scheduled evaluation, while adaptive tasks check pending work after their polling sleep. Could we replace “wakes on its schedule or after a notification” with something like “runs on its evaluation schedule or adaptive polling cadence and consumes pending dirty windows on its next evaluation”? The same notification-wakeup wording in `batching_mode.md` should be adjusted as well.

docs/contributor-guide/metasrv/overview.md
```diff
@@ -1,161 +1,95 @@
...
-- Election & High Availability, GreptimeDB is designed in a Leader-Follower architecture, only Leader nodes can write while Follower nodes can read, the number of Follower nodes is usually >= 1, and Follower nodes need to be able to switch to Leader quickly when Leader is not available.
-- Statistical data collection (reported via Heartbeats on each node), such as CPU, Load, number of Tables on the node, average/peak data read/write size, etc., can be used as the basis for distributed scheduling.
+Metasrv is the metadata and coordination service in a distributed GreptimeDB cluster. It does not sit on the data path. Its main responsibilities are:
+
+- storing Catalog, Schema, Table, Region, route, and node metadata;
+- choosing Datanodes for new Regions and maintaining table routes;
+- electing one Metasrv leader to coordinate metadata changes;
+- running recoverable procedures for DDL, Region migration, failover, and repartitioning;
+- tracking node leases and Region statistics through heartbeats;
+- notifying Frontends and Datanodes when cached metadata or Region state changes.
```
WenyXu: `MetasrvCacheInvalidator` broadcasts each `InvalidateCaches` instruction to the Frontend, Datanode, and Flownode mailbox channels. This bullet and the diagram omit Flownode and couple cache invalidation to Region-state notifications. Could we state that cache invalidations go to all three channels, while Region lifecycle instructions target Datanodes, and update the Chinese page as well?

docs/contributor-guide/metasrv/overview.md
```diff
@@ -1,161 +1,95 @@
...
-
-DynamoDB defines RCUs & WCUs (Read Capacity Units / Write Capacity Units), explaining that a RCU is a read request of 4KB data, and a WCU is a write request of 1KB data. When using RCU and WCU to describe workloads, it's easier to achieve performance measurability and get more informative resource preallocation because we can abstract different hardware capabilities as a combination of RCU and WCU.
+Datanodes maintain heartbeat streams to the Metasrv leader. Heartbeat requests report node identity, lease information, Region statistics, and other state used for placement and supervision. Responses carry control messages such as Region lifecycle instructions and cache invalidations.
 
-However, GreptimeDB still faces a more complex situation than DynamoDB, in particular, RCU doesn't fit to describe GreptimeDB's read workloads which require a lot of computation. We are working on that.
+Metasrv treats a heartbeat as a lease renewal, not merely as a metrics sample. Lease expiration is therefore part of failure detection and can lead to a Region failover procedure. Changes to heartbeat timing must remain consistent with the lease and supervision intervals.
```
WenyXu: The failover path is not driven by lease expiration. `DatanodeKeepLeaseHandler` renews a Datanode lease, while `RegionSupervisor` applies per-Region Phi Accrual detectors to heartbeat-arrival intervals and submits failover migrations only when Region failover is enabled. Could we separate these contracts here and link to the Region Failover guide for its enablement and storage prerequisites? The Chinese text has the same causal claim.

WenyXu (commented):
These comments clarify two Metasrv contracts that contributors must preserve: cache-invalidation fan-out and the separation between lease renewal and Region failure detection.

docs/contributor-guide/flownode/dataflow.md
killme2008: Fixed in f677db3d. `BatchingTask` no longer described as notification-woken. `dataflow.md` now says it runs on its evaluation schedule or adaptive polling cadence and collects pending dirty windows at that evaluation, and states explicitly that marking a window dirty does not wake the task. `batching_mode.md` is updated in the same way: the overview bullet, the execution-loop step (now distinguishing the scheduled sleep from the adaptive polling interval derived from window size and minimum refresh duration), and the "Task Wake-up" step, which is renamed to "Next Evaluation". Chinese pages updated to match.

docs/contributor-guide/metasrv/overview.md
killme2008: Fixed in f677db3d. Verified `MetasrvCacheInvalidator::broadcast` fans out to the Frontend, Datanode, and Flownode channels. The bullet is split into two: cache invalidations broadcast to all three, and Region lifecycle instructions target Datanodes. The diagram now reads `Region lifecycle instructions -> Datanode` and `cache invalidations -> Frontend / Datanode / Flownode`. Chinese page updated to match.

docs/contributor-guide/metasrv/overview.md
killme2008: Fixed in f677db3d. The causal claim is gone. Heartbeat Management now lists the two mechanisms separately: `DatanodeKeepLeaseHandler` renews the node lease (consumed by selectors and `/node-lease`), while `RegionSupervisor` keeps a per-Region Phi Accrual detector over heartbeat arrival intervals, with its verdict independent of lease expiry. It also states that a verdict submits a failover migration only when Region failover is enabled — off by default, requiring remote WAL unless `allow_region_failover_on_local_wal` is set — and that maintenance mode suppresses failover, with a link to the Region Failover guide. The Distributed Consensus sentence no longer attributes Region availability to leases. Chinese page updated to match.

killme2008:
`f677db3d` also contains four self-review fixes outside the commented lines:

- `contributor-guide/overview.md`: the previous sentence sent readers to `CONTRIBUTING.md` for build instructions, but that file has no prerequisites, toolchain, or build commands — the build guide is this repo's own Getting Started page. It now links Getting Started for building and scopes `CONTRIBUTING.md` to submission requirements.
- `getting-started.md`: restored `Git (optional)` and the 16GB/mold note. The build uses the `git2` crate and `open_repository` tolerates a missing repository, so the `git` CLI is only needed to clone; the memory/mold guidance is still valid and has no replacement elsewhere.
- `datanode/metric-engine.md`: restored the indirect-routing paragraph. `LogicalTableRouteValue` still stores only `physical_table_id`, so migrating a physical Region rewrites one physical route instead of every logical route. That is a current invariant, not design rationale.
- PR description: corrected two statements that did not match the diff — `user-guide/.../configuration.md` and `reference/sql/create.md` are changed, and `sidebars.ts` adds the new Memtable page.

`DOC_LANG=en pnpm check:links`, `DOC_LANG=zh pnpm check:links`, and `git diff --check` all pass.

One item is deliberately left out of this PR: `versioned_docs/version-1.1` and `version-1.2` (and their Chinese pages) still document `[region_engine.mito.memtable]` and `memtable.type = partition_tree`. `PartitionTreeMemtable` was removed in greptimedb `7840aa1bb4`, which shipped in v1.1.0, and unknown config keys are silently dropped, so a user following the v1.2 configuration page gets no error and no effect. Both released versions already have `sst_format`, so the Nightly wording backports cleanly. I will run the backport once this PR is approved.

docs/contributor-guide/datanode/memtable.md
```diff
@@ -0,0 +1,103 @@
...
+## Memtable implementations
+
+Mito selects a memtable implementation from the Region's SST format, primary-key encoding, and memtable options:
+
+```text
+memtable.type=bulk                                  -> BulkMemtable and flat SST
+flat SST format or sparse primary-key encoding      -> BulkMemtable
+primary_key SST, dense encoding, primary key exists -> TimeSeriesMemtable
+primary_key SST, dense encoding, no primary key     -> SimpleBulkMemtable
+```
```
evenyag, line 54: Maybe only mention BulkMemtable because we may deprecate other memetables later.

killme2008: Mostly done in 1f7eea83 — `BulkMemtable` is now clearly the subject of the page. The selection block drops from four rows to three and no longer names the legacy implementations individually, `### TimeSeriesMemtable` is gone, and what is left is one paragraph under `### Legacy implementations` marked as compatibility code that may be removed once `primary_key` is retired.

I stopped short of removing them entirely, for two reasons — happy to go all the way if you disagree.

`sst_format` is a documented table option in the current release and `primary_key` is a supported value, so regions created today can still land on `TimeSeriesMemtable` (or `SimpleBulkMemtable` when there is no primary key). This page has a "Constraints for changes" section aimed at people editing memtable code, and they will see `time_series.rs` and `simple_bulk_memtable.rs` in the tree. If the page does not mention them at all, the likely reading is that the page is stale rather than that the code is legacy.

The selection rule also needs them to mean anything: "flat format or sparse encoding requires `BulkMemtable`" is only informative while another outcome is possible.

When `primary_key` is actually retired, deleting that paragraph is a one-line diff in the same PR that removes the code. If you would rather not carry it until then, say so and I will cut it.

evenyag (approved):

PR #2785, opened 2026-08-25
docs: audit reference links and content

killme2008:
## What changed

- Audited English and Chinese reference documentation for Nightly, 1.0, 1.1, and 1.2, including links within reference pages and links between reference and user-guide pages.
- Corrected reference content against the applicable GreptimeDB source: stale CLI options, SQL and information-schema behavior, missing public functions and fields, unsafe pagination guidance, and other factual issues.
- Fixed semantic link destinations, source-path formatting, and published anchor compatibility.
- Tightened promotional or translation-heavy wording while preserving documented performance conclusions, recommended values, and compatibility claims that are not contradicted by the implementation.

The released-version changes follow the reviewed Nightly content where the behavior is shared. Version-specific differences remain intact: 1.0 keeps its RPC option names and older slow-query schema, 1.0/1.1 do not receive the 1.2 public API listener documentation, and each JSON function page reflects the functions registered by that release. The generated DataFusion function pages remain unchanged. Internal-only or experimental commands and endpoints were not added merely for implementation completeness.

## Scope

- Documentation versions: Nightly, 1.0, 1.1, and 1.2
- Languages: English and Chinese

Changes outside `reference/**` are limited to verified inbound links or destination anchors needed by the reference audit. No 0.x documentation is modified.

## Verification

- `DOC_LANG=en pnpm check:links`
- `DOC_LANG=zh pnpm check:links`
- `git diff --check`
- Manually reviewed the complete diff and confirmed that it does not modify 0.x documentation, DataFusion-generated pages, generated site output, or `pnpm-lock.yaml`.
- Cross-checked version-sensitive claims against GreptimeDB v1.0.2, v1.1.4, v1.2.0-beta.2, and current source. Reviewed source paragraphs, destination sections, and relevant overview pages for semantic link changes.

## Checklist

- [x] I verified the content against the applicable GreptimeDB version.
- [x] I updated the relevant documentation versions and languages, or explained why not.
- [x] I checked changed links and anchors.
- [x] No navigation update was required because the document structure did not change.

## Issues

Closes #2743

PR #2788, opened 2026-08-25
docs: fix broken hyperlink parsing caused by trailing punctuation

Austim-yang:
## What changed

A raw URL in the markdown file was not wrapped in angle brackets (<>). As a result, the Markdown parser incorrectly included the trailing period (。) into the hyperlink, which broke the link. Initially, I tried using angle brackets (<>), but it caused compilation errors, so i changed the raw URL to the standard inline link format (e.g., [URL](URL)). Only fix it in the Chinese version because the English version displays normally.

## Scope

- Documentation versions: every version
- Languages: Chinese

## Verification

checked by build in local.

## Checklist


Austim-yang:
Sorry. It has been discovered that an existing PR [#2785](https://github.com/GreptimeTeam/docs/pull/2785)has already fixed this issue.

Issue #2774, opened 2026-08-21
Update docs for fix(operator): reconcile internal events schema


sunng87:
A document change request is generated from https://github.com/GreptimeTeam/greptimedb/pull/8930

PR #2791, opened 2026-08-26
docs: document events and slow query table reconciliation

WenyXu:
## What changed

Documents the GreptimeDB #8930 behavior tracked by #2774 where it affects the `greptime_private.events` and `greptime_private.slow_queries` pages. The event and slow-query recorders create their system tables when absent and add missing columns to existing tables, even when automatic table creation is disabled.

## Scope

- Documentation versions: Nightly
- Languages: English, Chinese

## Verification

- `pnpm build`
- `DOC_LANG=zh pnpm build`
- `git diff --check`
- Reviewed the merged GreptimeDB #8930 implementation and the aggregate documentation diff.

## Checklist


docs/reference/sql/greptime-private/events.md
```diff
@@ -11,9 +11,12 @@ with Metasrv can additionally record operational event types. Event recording is
...
 operation succeeded.
 
-The table is created when the first event is recorded. If no event has been
-recorded, or event recording is disabled, querying it returns a table-not-found
-error. Configure recording in [Event recording](/user-guide/deployments-administration/configuration.md#event-recording).
+When the event recorder writes its first event, it creates the
+`greptime_private.events` system table. If an existing table is missing columns
+from the current event schema, the recorder adds them. It performs both actions
+even when automatic table creation is disabled. If no event has been recorded, or
+event recording is disabled, querying it returns a table-not-found error. Configure
```
fengjiachun: If this table was created while event recording was enabled, disabling recording later does not remove it, so queries still succeed. Could we narrow this to cases where the table has never been created—for example, because recording was disabled from startup? The same wording should be updated in the English and Chinese reference/overview pages.

WenyXu: Thanks, updated in 95512c26. The English and Chinese reference and overview pages now state that disabling event recording does not remove an existing table; table-not-found applies only when the table has never been created.

docs/reference/sql/greptime-private/events.md
```diff
@@ -11,9 +11,13 @@ with Metasrv can additionally record operational event types. Event recording is
...
 
-The table is created when the first event is recorded. If no event has been
-recorded, or event recording is disabled, querying it returns a table-not-found
-error. Configure recording in [Event recording](/user-guide/deployments-administration/configuration.md#event-recording).
+When the event recorder writes its first event, it creates the
+`greptime_private.events` system table. If an existing table is missing columns
+from the current event schema, the recorder adds them. It performs both actions
+even when automatic table creation is disabled. Disabling event recording does
+not remove an existing table. A query returns a table-not-found error only if the
+table has never been created, for example because event recording was disabled
```
fengjiachun: `greptime_private.events` can also be dropped. In that case, it has been created before, but queries return table-not-found until the recorder writes another event and recreates it. Could we say “only if the table does not currently exist” instead of “has never been created”? Please mirror the change in the English and Chinese reference/overview pages.

WenyXu: Thanks, updated in 2fdc8955. The English and Chinese reference and overview pages now say table-not-found applies when the table does not currently exist, including after it is dropped, and that the recorder recreates it on the next event write.

fengjiachun (approved):
LGTM

PR #2799, opened 2026-08-28
docs: update Flow user guide

discord9:
## What changed

- Refresh the Flow user guide, clarify batching versus scheduled full-query execution, and correct SQL examples and cleanup commands.
- Document missing-source deferral and the manual drop/recreate activation workflow, append-only experimental incremental reads, sink schema, and SQL/TQL `update_at` differences.
- Present Flow inspection commands and their purposes as tables in English and Chinese.
- Add the required evaluation interval to the non-windowed quick-start aggregate in all three versions.
- Backport release-valid guidance to v1.1 and v1.2. Keep their older mode-selection behavior: `EVAL INTERVAL` does not force batching there, while current Nightly does after GreptimeTeam/greptimedb#8878.
- Omit `SHOW FLOW STATUS` and `information_schema.flow_statistics` from v1.1 because those interfaces are unavailable in v1.1.4.

Source verification: GreptimeDB `27b0ca667634e132a13f29ff268f8b2472295bee` for current Nightly behavior, `v1.1.4`, and `v1.2.0-beta.2`.

## Scope

- Documentation versions: Nightly, v1.1, v1.2
- Languages: English, Chinese
- 24 Flow user-guide pages; no configuration, navigation, lockfile, or generated-file changes relative to the base branch.

## Verification

- `git diff --check origin/main...HEAD`
- `DOC_LANG=en pnpm check:links` (includes production build)
- `DOC_LANG=zh pnpm check:links` (includes production build)
- Source-based independent reviews of Nightly and release-specific behavior
- EN/ZH inspection-command parity and generated HTML table checks

## Checklist

- [x] I updated navigation when the document structure changed. No navigation change was required.


docs/user-guide/flow-computation/manage-flow.md
```diff
@@ -94,43 +83,39 @@ The sink table has the columns `sensor_id`, `loc`, `max_temp`, `time_window`, an
...
+[ EVAL INTERVAL <interval> ]
 [ COMMENT '<string>' ]
 [ WITH (<flow-option> = <value> [, ...]) ]
-AS 
+AS
 <SQL>;
 ```
 
+The clauses must appear in the order shown: `EXPIRE AFTER` comes before `EVAL INTERVAL`.
+`EVAL INTERVAL` schedules repeated evaluation of the full query. Scheduled SQL flows can use joins,
```
fengjiachun: `EVAL INTERVAL` does not schedule every SQL Flow in the current implementation. `determine_flow_type` still routes non-aggregate/non-`DISTINCT` SQL to the legacy streaming engine, and that engine explicitly ignores `eval_interval`; only batching SQL/TQL enters the scheduled loop. A projection or non-aggregate join can therefore accept this clause but still run with streaming semantics instead of periodic full-query evaluation. Please scope this statement to batching SQL (and explain how a query qualifies), or make `EVAL INTERVAL` force batching before documenting this contract.

docs/user-guide/flow-computation/manage-flow.md
```diff
@@ -142,31 +127,40 @@ COMMENT 'My first flow in GreptimeDB'
...
 For example, if the flow engine processes the aggregation at 10:00:00 and the `'1 hour'::INTERVAL` is set,
-any input data that arrive now with a time index older than 1 hour (before 09:00:00) will expire and be ignore.
-Only data timestamped from 09:00:00 onwards will be used in the aggregation and update to sink table.
+any input data that arrive now with a time index older than 1 hour (before 09:00:00) will expire and be ignored.
+Only data timestamped from 09:00:00 onwards will be used in the aggregation and to update the sink table.
+
+### Defer creation when a source is missing
+
+By default, creating a Flow fails if one of its source tables does not exist. Set
+`defer_on_missing_source` to `true` to persist a pending Flow instead of failing. The Flow is not scheduled while its
```
fengjiachun, line 155: This reads as though the pending Flow will become scheduled once the table appears. At this head, pending flows are persisted without flownode routes, source-table creation does not revisit them, and pending-to-active `CREATE OR REPLACE` is explicitly rejected. Please document the actual activation workflow (currently dropping and recreating the Flow after all sources exist) instead of implying automatic activation.

docs/user-guide/flow-computation/manage-flow.md
```diff
@@ -275,9 +254,26 @@ For more details on the behavior of the function,
...
+SHOW FLOWS;
+SHOW CREATE FLOW my_flow;
+SHOW FLOW STATUS LIKE 'my%';
+SELECT * FROM information_schema.flows;
+SELECT * FROM information_schema.flow_statistics;
+```
+
+`SHOW FLOWS` lists flows, `SHOW CREATE FLOW` returns a Flow definition, and `SHOW FLOW STATUS` returns runtime
+statistics. The `information_schema` tables provide definition and statistics details. Runtime fields can initially be
+`NULL`, and values can lag behind the latest state in distributed deployments.
```
killme2008: Make this as a table would be more clear

killme2008 (approved):
LGTM

Don't forget to backport them into v1.2 and v1.1 if possible.

docs/user-guide/flow-computation/manage-flow.md
discord9: Updated in 79e5cfd2: the inspection commands and their purposes are now tables in both English and Chinese. I also backported the guide to v1.1 and v1.2 with version-specific behavior: v1.1.4 lists only SHOW FLOWS, SHOW CREATE FLOW, and information_schema.flows; v1.2 retains all five interfaces. Both locale production builds and strict link checks pass.

docs/user-guide/flow-computation/manage-flow.md
discord9: Thanks — this was correct at the reviewed implementation. Since then, GreptimeTeam/greptimedb#8878 (ad7b0ace64a) made EVAL INTERVAL select batching regardless of SQL shape and reject instant-TTL sources with the interval. The updated Nightly guide follows that behavior and distinguishes scheduled evaluations from unfiltered full-query execution. The v1.1/v1.2 backports retain the older routing: Aggregate/Distinct SQL and TQL select batching, instant-TTL sources select streaming, and streaming ignores EVAL INTERVAL. Updated EN/ZH in 79e5cfd2; both strict locale checks pass.

docs/user-guide/flow-computation/manage-flow.md
discord9: Fixed in 79e5cfd2 in English and Chinese, including the v1.1/v1.2 backports. The guide now explicitly says that creating the missing source tables does not activate a pending Flow and CREATE OR REPLACE cannot activate it. A separate SQL example shows DROP FLOW followed by recreation after all source tables exist. Both locale builds and strict link checks pass.

PR #2803, opened 2026-08-30
docs: add single-region DR solution to disaster recovery overview

fengjiachun:
## What changed

The disaster recovery overview jumped from GreptimeDB Standalone straight to the cross-region topologies. Missing was the tier most production clusters actually run: a single cluster inside one region, tolerating the loss of a node or of a single AZ through Region Failover.

This adds that section to the overview in both locales, plus a matching row in the solution comparison table.

## How it is written

The section follows the shape of [DR Solution Based on Cross-Region Deployment in a Single Cluster](https://docs.greptime.cn/1.2/user-guide/deployments-administration/disaster-recovery/dr-solution-based-on-cross-region-deployment-in-single-cluster/), so the two read as one family:

- Same AZ definition as that page: a data center, or a compartment of one. The section therefore covers both a single data center and multiple AZs in one region.
- The same `Latencies:` and `Supports High Availability:` lists.
- Per-component roles: stateless Frontend replicas, Metasrv replicas plus the external metadata backend it does not replicate for you, Datanode with Region Failover, and Kafka plus object storage as the state that outlives the node.
- The conditions the targets depend on: Region Failover is disabled by default, the surviving Datanodes need spare capacity, and the recovery time is dominated by WAL replay.

Facts are taken from the existing pages on Region Failover, WAL and metadata management. The latency list does not repeat the example millisecond values from the cross-region page, since there are no measured single-region numbers to quote.

## Diagram

New `static/Single-region-single-cluster.svg`, drawn in the same style as `static/img/active-active-forwarding.svg` (Arial, `#6f6fff`, `title`/`desc` for accessibility) and shared by both locales like the other DR images.

## Scope

Nightly only: `docs/` and the Chinese `current/` tree. Backporting to v1.2 can follow separately.

## Checks

- `DOC_LANG=en pnpm check:links` passes
- `DOC_LANG=zh pnpm check:links` passes, including the new image path and the Chinese anchors
- `git diff --check` clean; no lockfile or generated files touched

### DIFF
--- a/docs/user-guide/deployments-administration/disaster-recovery/overview.md
+++ b/docs/user-guide/deployments-administration/disaster-recovery/overview.md
@@ -1,6 +1,6 @@
 ---
-keywords: [disaster recovery, GreptimeDB, DR solutions, backup and restore, active-active failover, cross-region deployment, RTO, RPO]
-description: Overview of disaster recovery (DR) solutions in GreptimeDB, including basic concepts, component architecture, and various DR solutions such as standalone, active-active failover, cross-region deployment, and backup & restore.
+keywords: [disaster recovery, GreptimeDB, DR solutions, backup and restore, single-region deployment, region failover, active-active failover, cross-region deployment, RTO, RPO]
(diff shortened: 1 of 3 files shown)


fengjiachun:
Superseded by #2804 — the branch was renamed to follow the repository's `docs/<topic>` convention.

PR #2804, opened 2026-08-30
docs: add single-region DR solution to disaster recovery overview

fengjiachun:
Adds the tier missing between Standalone and the cross-region topologies: a single cluster in one region, surviving the loss of a node or of a single AZ through Region Failover.

Written to match [the cross-region page](https://docs.greptime.cn/1.2/user-guide/deployments-administration/disaster-recovery/dr-solution-based-on-cross-region-deployment-in-single-cluster/) — same AZ definition (a data center, or a compartment of one), same `Latencies:` and `Supports High Availability:` lists — plus a row in the comparison table and a new diagram. Both locales, Nightly only.

The latency list carries no millisecond values: the ones on the cross-region page are labelled examples, and there are no measured single-region numbers to quote.

`DOC_LANG=en` and `DOC_LANG=zh` `pnpm check:links` both pass.


docs/user-guide/deployments-administration/disaster-recovery/overview.md
```diff
@@ -126,6 +155,7 @@ By comparing these DR solutions, you can decide on the final option based on the
 |     DR solution | Error Tolerance Objective |  RPO | RTO | TCO | Scenarios | Remote WAL & Object Storage | Notes |
 | ------------- | ------------------------- | ----- | ----- | ----- | ---------------- | --------- | --------|
 |  DR solution for Standalone| Single-Region | Backup Interval | Minute or Hour level | Low | Low requirements for availability and reliability in small scenarios |  Optional | |
+|  DR solution based on single-region deployment in a single cluster | Single-Region, node and AZ level | 0 | Minute level | Medium | The common production baseline for a cluster that lives in one region | Required | Region Failover is disabled by default |
```
killme2008: Let's change the RTO from `minute-level` to `second- or minute-level`, depending on feasibility, and update the RTO for the `DR solution based on cross-region deployment in a single cluster` accordingly.

killme2008 (approved):

docs/user-guide/deployments-administration/disaster-recovery/overview.md
fengjiachun: Done in 38642d9 — both the prose and the comparison table now say second- or minute-level, for this solution and for the cross-region one.

Left the Standalone entry alone: its remote-WAL topology still says an RTO in minutes and the table says minute or hour level, since recovery there waits on a human restarting the instance. Say the word if you want that one widened too.

PR #2806, opened 2026-08-31
chore: backport iceberg docs to 1.2

sunng87:
## What changed

Follow up #2748 and backport it to 1.2

## Scope

- Documentation versions: 
- Languages: 

## Checklist

- [ ] I verified the content against the applicable GreptimeDB version.
- [ ] I updated the relevant documentation versions and languages, or explained why not.
- [ ] I checked changed links and anchors.

### DIFF
--- a/versioned_docs/version-1.2/enterprise/iceberg.md
+++ b/versioned_docs/version-1.2/enterprise/iceberg.md
@@ -1,0 +1,321 @@
+---
+keywords: [Iceberg, REST catalog, pyiceberg, Spark, parquet, data export, open table format]
+description: Read GreptimeDB tables through an Apache Iceberg REST catalog with pyiceberg, Spark, Trino, DuckDB, and other engines — no connector or data copy required.
+---
+
+# Iceberg Export
+
+GreptimeDB Enterprise can expose its tables through an [Apache Iceberg](https://iceberg.apache.org/) REST catalog,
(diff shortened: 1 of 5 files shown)


versioned_docs/version-1.2/enterprise/iceberg.md
```diff
@@ -0,0 +1,321 @@
...
+GreptimeDB publishes Iceberg metadata on flush, so the export of recently written data appears after the next flush
+(or compaction). To expose freshly-ingested rows immediately, flush the table manually:
+
+```sql
+-- via the MySQL or PostgreSQL protocol
+admin flush_table('your_table');
+```
+
+The metadata is published asynchronously; the table becomes queryable through the REST catalog shortly after the
+flush returns. Existing, already-flushed tables are exported automatically.
```
fengjiachun, line 143: This is not automatic for tables whose SSTs were already flushed before the plugin was enabled: the hook only observes subsequent manifest updates, and there is no startup bootstrap to publish those existing files. Such a table can therefore remain absent from the Iceberg catalog until a later manifest change or an explicit rebuild, which also conflicts with the rebuild case documented below. Could we direct operators to run the rebuild endpoint for pre-existing tables and make the same correction in the Chinese page?

killme2008 (approved):

versioned_docs/version-1.2/enterprise/iceberg.md
sunng87: will address this in a separated patch

PR #2814, opened 2026-09-02
docs: shorten the batching mode title and rework the flownode diagrams

killme2008:
## What changed

Rename the Flownode batching mode page to **Batching Mode**. The old heading
`Flownode Batching Mode Developer Guide` also drove the sidebar label, which was
too long for the flownode section. Cross-references in `dataflow.md` and
`overview.md` follow the new name.

Replace `batching_mode_arch.png` with an inline mermaid diagram, matching the
plugin added in #2801 and the existing usage in `enterprise/iceberg.md`. The
rendered diagram goes from 633x1582 to 387x608, and mermaid picks up the site's
configured light and dark themes instead of shipping a light-only raster.

The old image also had errors, which the new diagram fixes:

- It drew `BatchingEngine` and `Database` as decision diamonds. Neither is a decision.
- Its `Notify` edge implied that a source-table write drives the task. #2784
  corrected this: marking a window dirty does not wake the task, which runs on its
  evaluation schedule or adaptive polling cadence.
- `Database` is vague. The task encodes an `INSERT INTO sink SELECT` plan and sends
  it through `FrontendClient`, so the node is now `Frontend`.
- Frontend scans the source table while executing that plan. The old image had no
  source-table read edge.

`static/batching_mode_arch.png` is kept rather than deleted, in case the URL is
referenced outside this repository.

### Flow mode selection

`dataflow.md` listed the mode-selection rules without their precedence, so a reader
could conclude that an aggregating Flow on a source table with `ttl = 'instant'` runs
in batching mode. `determine_flow_type` checks the source tables first: a missing
source table needs `defer_on_missing_source` and yields a pending batching Flow, an
instant-TTL source forces streaming, and only then are TQL and `Aggregate`/`Distinct`
considered. The paragraph is replaced with a flowchart in that order, plus an explicit
sentence about the instant-TTL consequence.

v1.0 gets a shorter diagram. `defer_on_missing_source` does not exist in v1.0.2, where
an instant-TTL source returns streaming directly from the source-table loop.

## Scope

- Documentation versions: Nightly, 1.2, 1.1, 1.0
- Languages: English, Chinese

## Verification

- `git diff --check`
- Rendered every built page in headless Chrome for all four versions and both
  locales: the H1 is `Batching Mode` / `批处理模式`, and each page emits exactly one
  `aria-roledescription="flowchart"` SVG.
- Verified the mermaid blocks are byte-identical across the versions that share them,
  and that the v1.0 flow-selection diagram omits the `defer_on_missing_source` branch.
- Verified the diagram against GreptimeDB `main`: `gen_insert_plan_unlocked` builds a
  `LogicalPlan::Dml(WriteOp::Insert)` over the sink table, `encode_insert_plan_request`
  wraps it as `api::v1::InsertIntoPlan`, and it is sent through `FrontendClient`, so
  Frontend performs both the source scan and the sink write.
- No lockfile or `package.json` change.

## Checklist

- [x] I updated navigation when the document structure changed. No sidebar file changes
      are needed: the entry is the doc id, and its label comes from the H1.


docs/contributor-guide/flownode/batching_mode.md
```diff
@@ -22,7 +22,16 @@ The core idea is to:
...
 The batching mode consists of several key components that work together to achieve this continuous aggregation. As shown in the diagram below:
 
-![batching mode architecture](/batching_mode_arch.png)
+```mermaid
+flowchart TB
+    SRC["Source table"] -->|"write"| ENG["BatchingEngine"]
+    ENG -->|"mark affected windows"| DTW("DirtyTimeWindows<br/>pending ranges per task")
+    DTW -->|"read at next evaluation"| RUN["BatchingTask<br/>schedule or polling cadence"]
+    RUN -.->|"clear processed"| DTW
+    RUN -->|"INSERT INTO sink SELECT<br/>with time predicates"| FE["Frontend"]
```
discord9: This diagram currently makes `DirtyTimeWindows` and `with time predicates` look mandatory for every batching Flow. TQL and evaluation-interval SQL without a usable time-window expression take an unfiltered full-query path, while windowed SQL uses dirty-window predicates. Could we label this path “when windowed” or draw the unfiltered path separately before applying the diagram to every version?

versioned_docs/version-1.0/contributor-guide/flownode/dataflow.md
```diff
@@ -22,7 +35,7 @@ Batching mode reuses GreptimeDB's query engine instead of maintaining an operato
 4. The query result is inserted into the sink table, updating the materialized result for windows that were evaluated.
 5. Successfully processed windows are removed from the dirty set. Failed work remains available for a later evaluation.
 
-Flows with an evaluation interval but without a time-window expression run the complete query on each scheduled evaluation. This path also lets Flow use query-engine features that the streaming renderer does not implement. See [Flownode Batching Mode Developer Guide](./batching_mode.md) for the task and dirty-window components.
+Flows with an evaluation interval but without a time-window expression run the complete query on each scheduled evaluation. This path also lets Flow use query-engine features that the streaming renderer does not implement. See [Batching Mode](./batching_mode.md) for the task and dirty-window components.
```
discord9: This differs in v1.0: no-window SQL first checks whether dirty input exists and skips execution when clean. Could this be release-specific, for example: “When input is marked dirty, a batching Flow with an evaluation interval but no time-window expression executes the complete query rather than a time-filtered query”?

docs/contributor-guide/flownode/batching_mode.md
```diff
@@ -3,7 +3,7 @@ keywords: [batching mode, flow management, Flownode components, Flownode limitat
 description: Overview of Flownode's batching mode, the active execution mode for continuous data aggregation, including its architecture and query execution flow.
 ---
 
-# Flownode Batching Mode Developer Guide
+# Batching Mode
```
discord9, line 6: This H1 was already published, so shortening it changes the existing fragment URL. Per `CONTRIBUTING.md`, please add an `AnchorAlias` immediately before each renamed EN/ZH H1 to preserve the previous anchor across Nightly and versions 1.0–1.2.

docs/contributor-guide/flownode/batching_mode.md
killme2008: Fixed in 5748d9a8, taking your first option. The `BatchingTask` node now has two labelled edges into `Frontend`:

- `windowed: INSERT INTO sink SELECT with time predicates`
- `TQL or unprunable plan: unfiltered full query`

Confirmed against `gen_query_with_time_window`: only `(Some((Some(l), Some(u))), QueryType::Sql)` takes the dirty-window predicate path, and the catch-all arm calls `gen_unfiltered_plan_info(... QueryCoverage::UnfilteredFull)`. I also picked up the point in the code comment there — on that path the dirty set is only a scheduling signal — and stated it in `dataflow.md`.

While rewriting that sentence I noticed it named only evaluation-interval SQL and omitted TQL, so it now reads "TQL Flows, and evaluation-interval Flows whose plan cannot be pruned safely by dirty windows, run the complete query rather than a time-filtered one."

The diagram grows from 387x541 to 541x608 in width only; height is unchanged, so the page does not get taller. Applied to Nightly and 1.0 through 1.2 in both locales.

versioned_docs/version-1.0/contributor-guide/flownode/dataflow.md
killme2008: Fixed in 5748d9a8. You are right, and the difference is sharper than the sentence implied, so v1.0 now carries its own wording.

In `v1.0.2` the `(None, QueryType::Sql)` arm reads the dirty set, clears it, and returns `Ok(None)` when it was empty, so a clean evaluation is skipped entirely. The catch-all arm that TQL falls into does not do that: it comments `clean for tql have no use for time window`, clears the set, and always builds the full query. So the skip applies to no-window SQL only, not to TQL.

v1.0 now reads: "TQL Flows, and SQL Flows with an evaluation interval but without a time-window expression, run the complete query rather than a time-filtered one. For the SQL case, an evaluation that finds nothing dirty is skipped."

Nightly, 1.1 and 1.2 get different wording, since `v1.1.4` onward replaced that arm with an unconditional `drain_dirty_windows_signal()` plus `QueryCoverage::UnfilteredFull` and no longer skips on a clean evaluation.

docs/contributor-guide/flownode/batching_mode.md
killme2008: I checked this before renaming, and I do not think an alias is needed here: Docusaurus does not give the page title an anchor, so no fragment was ever published for it.

Three things I verified on the built site:

- The page title renders as a bare `<h1>` with no `id`, while `<h2>` on the same page gets `class="anchor anchorWithStickyNavbar_..." id="..."`. Docusaurus only assigns heading ids to `h2` through `h6`.
- Sampling 200 built pages across the site, none has an `id` on `h1`.
- A repo-wide search for `#flownode-batching-mode-developer-guide` and its Chinese equivalent returns nothing.

The page URL comes from the file path, and the rename only touched the heading text, so `/contributor-guide/flownode/batching_mode/` is unchanged. `CONTRIBUTING.md` asks us to preserve fragments that have already been published; adding an `AnchorAlias` here would register one for a fragment that never existed.

Happy to add it if you know of an external link that depends on the old fragment, or if you were pointing at something other than the URL, such as search indexing.

discord9 (approved):
The reported Flow execution-path and v1.0 behavior issues are fixed. I also verified that the renamed H1s do not generate Docusaurus fragment IDs, so no AnchorAlias is needed. The release-specific EN/ZH content is aligned, and all checks pass.

PR #2815, opened 2026-09-03
docs: document widening the time index timestamp unit via ALTER TABLE

sunng87:
## What's changed

GreptimeDB now allows `ALTER TABLE ... MODIFY COLUMN` to **widen the time index column's timestamp unit** to a finer one (e.g. `TIMESTAMP` milliseconds → `TIMESTAMP_US` microseconds) on non-metric-engine tables (greptimedb #8894). This PR updates the docs that previously stated the time index could not be modified at all:

- **`reference/sql/alter.md`** (Modify column type): field columns still can't be tags and must be nullable; the time index is now the documented exception — widening is allowed (with example), is lossless (existing data is read in the new unit without rewriting), while narrowing, non-timestamp targets, and metric engine tables remain unsupported.
- **`user-guide/.../basic-table-operations.md`**: the warning now says the TIME INDEX *column* can't be moved to another column, but its timestamp unit can be widened via `MODIFY COLUMN` on non-metric-engine tables, linking to the ALTER reference anchor.
- **`faq-and-others/faq.md`**: the "Can I change a column's type after table creation?" answer drops "or time index" and adds the widening exception with an example.

English (`docs/`, nightly) and Chinese (`i18n/zh/.../current/`) updated together with equivalent meaning. No released-version changes since the feature is unreleased.

## Checklist

- [x] `DOC_LANG=en pnpm check:links`
- [x] `DOC_LANG=zh pnpm check:links` (caught and fixed a zh anchor — the link now uses the Chinese heading anchor `#修改列类型`)
- [x] `git diff --check` clean; no lockfile or generated-file changes

### DIFF
--- a/docs/faq-and-others/faq.md
+++ b/docs/faq-and-others/faq.md
@@ -121,7 +121,13 @@ Yes. Use `ALTER TABLE ... MODIFY COLUMN` to change a field column's data type:
 ALTER TABLE monitor MODIFY COLUMN load_15 STRING;
 ```
 
-The column must be a field (not a tag or time index) and must be nullable, so that values that cannot be cast return `NULL` instead of failing.
+The column must be a field (not a tag) and must be nullable, so that values that cannot be cast return `NULL` instead of failing.
+
+The time index column is the exception: its timestamp unit can be widened to a finer one (for example, from milliseconds to microseconds) on tables that are not using the metric engine:
+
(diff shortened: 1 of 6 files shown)


fengjiachun (approved):
LGTM

killme2008 (approved):

PR #2846, opened 2026-09-09
docs: close v1.2 documentation gaps

### LINKED ISSUE #2595: Update docs for feat(flow): add information_schema.flow_statistics and SHOW FLOW STATUS (#7987)
A document change request is generated from https://github.com/GreptimeTeam/greptimedb/pull/8392

killme2008:
## What changed

Gaps found while auditing v1.2 against `v1.1.0..v1.2.0`.

Version pointers:

- The five `reference/command-lines/*` pages still linked config examples at `blob/v1.2.0-beta.2/`. Now `blob/v1.2.0/`, matching how v1.1 pins `v1.1.4`.

Backported to v1.2 (shipped in v1.2.0, previously documented only in Nightly):

- `admin_function` event type and the `actor` column: standalone event-type list, the Frontend list, the Event actor and ADMIN function column sections, and the two matching query recipes.
- `ALTER TABLE ... SET 'skip_wal'='true'` in the ALTER reference.

New in both versions:

- `prom_store.flow_notification_queue_capacity`, default 1024, batching mode only.
- `information_schema.flow_statistics` reference page and a `SHOW FLOW STATUS` section, wired into the sidebars and the information-schema overview. Closes #2595.

Corrections in both versions:

- `prom_validation_mode` and `experimental_enable_prometheus_native_histogram` were documented under `[prom_store]`. Both are `[http]` options upstream, so the TOML sample and the option table were wrong in all four files; configuring them as documented has no effect.
- `skip_wal` now records its applicability: physical tables on the `mito` or `metric` engine, and it must be set on its own.
- `flow_statistics` notes that `start_time` and `uptime_seconds` are always `NULL` in a cluster.

Not included: pipeline `type: json2` and its type hints, which also need backporting to v1.2. Left out to keep this reviewable.

## Scope

- Documentation versions: Nightly, 1.2
- Languages: English, Chinese

## Verification

Claims checked against the `v1.2.0` tag:

- `actor` at `src/common/event-recorder/src/event_table.rs:93`, `ADMIN_FUNCTION_EVENT_TYPE` at `src/operator/src/statement/admin/event.rs:34`
- `skip_wal` enable-only, `mito`/`metric` engines, physical tables only, and rejection of combined table options, per `src/common/meta/src/ddl/alter_table.rs:150-176`; rejection of `= 'false'` and `UNSET` per `tests/cases/standalone/common/skip_wal.result`
- `http.prom_validation_mode` and `http.experimental_enable_prometheus_native_histogram` per `config/config.md` and `config/frontend.example.toml`
- `flow_notification_queue_capacity` default and the drop metric `greptime_prom_store_flow_notification_dropped_total` at `src/servers/src/metrics.rs:181`
- `flow_statistics` column types and nullability at `src/catalog/src/system_schema/information_schema/flow_statistics.rs`; `SHOW FLOW STATUS` sorting and projection at `src/query/src/sql.rs:1127`; the cluster NULL behavior follows from `FlowStateValue` not carrying `start_time_map` in `src/common/meta/src/key/flow/flow_state.rs`

Commands:

- `pnpm build`
- `DOC_LANG=zh pnpm build`
- `git diff --check`

Both builds report broken links only on `migrate-from-influxdb` and `migrate-from-prometheus`, which this PR does not touch.

## Checklist


killme2008:
Addressed the Copilot review, plus two fixes found while checking it.

**`skip_wal` scope — valid, fixed.** Verified against `v1.2.0`: `src/common/meta/src/ddl/alter_table.rs:150-176` rejects the option on engines other than `mito`/`metric`, on logical tables, and when combined with other table options in the same statement. Added all three constraints. Nightly had the same omission, so it was updated too, keeping the four files identical.

**Chinese link spacing — not changed.** `请参阅[事件发起人]` was copied verbatim from the Nightly page, and the repository uses both `请参阅[` and `请参阅 [` (4 and 3 occurrences in this directory alone). Editing only the v1.2 copy would make the two versions diverge for a formatting preference the repository has never settled. Worth a separate pass if we want one convention.

Two unrelated problems surfaced while verifying the above, fixed in the same commit since they sit in the files this PR already touches:

- `prom_validation_mode` and `experimental_enable_prometheus_native_histogram` were documented under `[prom_store]` in all four `configuration.md` files. Upstream both are `[http]` options (`config/config.md`, `config/frontend.example.toml`), so following the docs would have had no effect.
- `flow_statistics` now states that `start_time` and `uptime_seconds` are always `NULL` in a cluster — `FlowStateValue` does not carry `start_time_map`, so the conversion drops it.

WenyXu (approved):

PR #2853, opened 2026-09-09
docs: fix export-import-v2 local filesystem and S3 guidance for v1.2

fengjiachun:
## What

Corrections to the Export/Import V2 page, based on an end-to-end run of the page against `greptime/greptimedb:v1.2.0` (commit `29b743ab`): standalone with a local filesystem snapshot, and a MinIO container for the object storage path. The rest of the page matched observed behavior and is left untouched.

Applies to `docs/` and `versioned_docs/version-1.2/`, plus the matching Chinese pages. The English pair and the Chinese pair are each kept identical.

### 1. Quick start fails on a default v1.2 configuration (copy root sandbox)

The first command in "Quick start with local filesystem" fails as written:

```bash
greptime cli data export-v2 create \
  --addr 127.0.0.1:4000 \
  --to file:///tmp/greptime-snapshots/demo
```

Server-side error (HTTP 400 on `/v1/sql`):

```text
Local filesystem path '/tmp/greptime-snapshots/demo/data/greptime_private/1/' is outside the configured copy root or is unsafe: absolute path is outside the configured copy root; use a path relative to the copy root or use S3, OSS, GCS, or AzBlob
```

`export-v2` writes data files through `COPY DATABASE ... TO`, and since v1.2 local paths in `COPY` are sandboxed to `storage.copy_root` (default `<storage.data_home>/copy`). Both fixes were verified: setting `[storage] copy_root = "/tmp/greptime-snapshots"` makes the documented command succeed unchanged, and using a path under the default copy root works too. Object storage targets are not subject to the sandbox (the server writes `s3://` directly), which the page now states next to its existing "use object storage for remote/distributed deployments" advice.

The page now:

- States the copy root constraint in Prerequisites and links to [Migrate Local SQL File Access](https://docs.greptime.com/user-guide/deployments-administration/migrate-local-sql-file-access).
- Adds the `storage.copy_root` snippet (and a `mkdir -p`) to the quick start so every `file:///tmp/greptime-snapshots/...` example on the page runs as written, with the default-copy-root path as the alternative.
- Adds a Troubleshooting entry with the error above. The failed run leaves a partial snapshot on disk that `export-v2 verify` reports as a failed chunk, so the entry also says how to resume or recreate it.

### 2. `--format` was undocumented

`export-v2 create --format <parquet|csv|json>` (default `parquet`) was not mentioned anywhere. CSV was verified end to end: the export produces `.csv` data files, and `import-v2` picks the format up from the manifest and issues `COPY ... WITH (FORMAT='csv')` on its own. Added a short section saying so, including that the format does not need to be repeated at import time.

### 3. Default schema set includes `greptime_private`

With no `--schemas`, v1.2.0 exports `["greptime_private", "metrics", "public"]`, excluding only `information_schema`, and `greptime_private` is restored on import. The page now states this and points to `--schemas` for excluding it.

### 4. Parent directory must exist

For `file://` snapshots the parent directory has to exist beforehand. Added one sentence in Prerequisites and the `mkdir -p` step in the quick start.

## Object storage (S3 / MinIO)

The documented S3/MinIO commands work as written: create, verify, list, import, and delete all succeed with the same option set, and a cross-instance restore matches row counts. Two statements around them do not hold.

### 5. `--s3-bucket` and `--s3-root` do not affect the snapshot location

The page told readers to keep `--s3-bucket` consistent with the URI and, under `NoSuchBucket`, to check that the URI bucket matches `--s3-bucket`. On v1.2.0 the bucket in the URI is the only thing that matters:

| Experiment | Command | Result |
| --- | --- | --- |
| Mismatched bucket, read | URI `s3://greptime/...` + `--s3-bucket wrongbucket`, `verify` | `Snapshot is valid.` |
| Mismatched bucket, write | same options, `create` | Succeeds; data lands in `greptime`, `wrongbucket` is never created or accessed |
| No `--s3-bucket` at all | `--s3` + `--s3-region` + credentials + `--s3-endpoint` | `Snapshot is valid.` |
| Reverse | URI `s3://nosuchbucket/...` + `--s3-bucket greptime` | `Snapshot not found at 's3://nosuchbucket/...'` |
| `--s3-root /myprefix` | URI `s3://greptime/snapshots/rootTest` | Snapshot still at `snapshots/rootTest/`; `myprefix` never appears |

The only options actually enforced are `--s3` (`s3:// requires --s3 and related options`) and `--s3-region` (`S3 region must be set when --s3 is enabled`), which the page already had right.

Changed: the note no longer asks readers to keep `--s3-bucket` consistent and says the URI is authoritative; `--s3-bucket` is no longer listed as required (Prerequisites, the AWS paragraph, the options table); the `NoSuchBucket` troubleshooting item points at the URI instead. The examples keep `--s3-bucket` as they were, since they work.

### 6. A missing `--s3-endpoint` hangs instead of failing

Against MinIO without `--s3-endpoint`, the CLI does not fail fast; it sends requests to `https://s3.us-west-2.amazonaws.com/...`, gets `301`, and keeps retrying (still running after 20 s, killed by hand). The troubleshooting item already said to check the endpoint; it now describes this symptom so readers recognize it.

## Checks

- `DOC_LANG=en pnpm check:links` and `DOC_LANG=zh pnpm check:links` pass, including the new cross-links in the versioned copies.
- `git diff --check` clean; no lockfile or generated files touched.

### DIFF
--- a/docs/user-guide/deployments-administration/disaster-recovery/export-import-v2.md
+++ b/docs/user-guide/deployments-administration/disaster-recovery/export-import-v2.md
@@ -25,15 +25,15 @@ Before using Export/Import V2, make sure that:
 - You have a `greptime` binary that includes the `cli data export-v2` and `cli data import-v2` commands.
 - The snapshot storage location is readable and writable by both the CLI client and the GreptimeDB server.
 
-For remote object storage, explicitly enable one supported backend and provide the backend options. Export/Import V2 supports S3-compatible storage, Alibaba Cloud OSS, Google Cloud Storage, and Azure Blob Storage. For example, use `--s3` with `--s3-bucket` and `--s3-region` for S3-compatible storage.
+For remote object storage, explicitly enable one supported backend and provide the backend options. Export/Import V2 supports S3-compatible storage, Alibaba Cloud OSS, Google Cloud Storage, and Azure Blob Storage. For example, use `--s3` with `--s3-region` for S3-compatible storage.
 
 For `file://` snapshots, the path must be accessible from the GreptimeDB server as well as the CLI client. This usually means running the CLI on the same host as a standalone server, or mounting the same filesystem path into the GreptimeDB server. The parent directory of the snapshot must exist before you create the snapshot. For remote, Kubernetes, or distributed deployments, use object storage such as S3 or MinIO instead of a local `file://` path.
 
-The GreptimeDB server writes snapshot data files with `COPY DATABASE ... TO`, so a `file://` snapshot path must also be inside the server's copy root, which is `storage.copy_root` and defaults to `<storage.data_home>/copy`. The server rejects a path outside the copy root. See [Migrate Local SQL File Access](/user-guide/deployments-administration/migrate-local-sql-file-access.md) for details.
+The GreptimeDB server writes snapshot data files with `COPY DATABASE ... TO`, so a `file://` snapshot path must also be inside the server's copy root, which is `storage.copy_root` and defaults to `<storage.data_home>/copy`. The server rejects a path outside the copy root. Object storage snapshots are not subject to the copy root, because the server writes them directly to the object store. See [Migrate Local SQL File Access](/user-guide/deployments-administration/migrate-local-sql-file-access.md) for details.
 
 :::note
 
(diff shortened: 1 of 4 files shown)


fengjiachun:
@codex review

killme2008 (approved):

PR #2869, opened 2026-09-12
docs: redraw the disaster recovery diagrams as SVG

fengjiachun:
Twelve PNG diagrams in the DR docs were still in an older visual style, while the same docset had already moved to hand-authored SVG — `img/active-active-forwarding.svg` (#2689) and the concepts pages. This redraws all twelve in that style: vector, Arial, the `#6f6fff` palette, `title` and `desc` for accessibility.

## Two kinds of change

**The eight overview diagrams** keep their content unchanged — same boxes, same labels, same numbers. Only the drawing changes.

**The four cross-region diagrams** are remodelled to use the element set of `Single-region-single-cluster.svg`, so the overview and the cross-region page read as one family: Metasrv and Frontend over AZ frames holding one Datanode per DC, a Region Failover, and one shared Remote WAL / object storage / metadata backend tier.

What that trades away, deliberately:

- leader and follower roles on the Metasrv and WAL broker pills — the 2-2-1 majority is now carried by the per-region `Metasrv × N` counts
- the split between `wal metasrv` and `wal brokers`, and the per-AZ broker placement — folded into the single `Remote WAL` element
- two Datanodes per AZ — one per DC now, as in the single-region diagram
- `T1-1` style table shard ids — now `Region N`, with colour still grouping the regions of one table

What it adds: the object storage and metadata backend tiers, which the page's own prose requires but the diagrams never showed, and a Region Failover, which none of them showed.

## Errors found in the current diagrams

- `BR-explain`: "dowload" — fixed.
- `Cross-region-single-cluster`: `Region 2-2` on both Datanode 2 and Datanode 3 — fixed. Table 2 sits on Datanodes 1, 2, 3 and 5, and now runs 2-1 to 2-4 in Datanode order, the way table 1 runs 1-1 to 1-5 and table 3 runs 3-1 to 3-4.
- `BR-DR`: `Region 1-2` twice inside Cluster 2's Datanode — fixed. Cluster 1 holds Region 1-1 on Datanode 1 and Region 1-2 on Datanode 2, so the restored Datanode carries 1-1 and 1-2.

Correct them in review if the intended distribution was different.

## Scope

Nightly and 1.2. Version 1.2 referenced the same twelve PNGs and its cross-region page is word for word the nightly one, so that swap is a reference change only. The PNGs stay in `static/` for 1.1 and older, which still reference them.

Chinese variants (`*.zh.svg`, the pattern the August concepts diagrams use) are not part of this PR — both locales share the English SVGs, exactly as they share the PNGs today.

## Checks

- `DOC_LANG=en pnpm check:links` passes
- `DOC_LANG=zh pnpm check:links` passes
- `git diff --check` clean; no lockfile or generated files touched

### DIFF
--- a/static/BR-explain.svg
+++ b/static/BR-explain.svg
@@ -1,0 +1,73 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="1200" height="700" viewBox="0 0 1200 700" role="img" aria-labelledby="title description">
+  <title id="title">Backup and restore between a GreptimeDB cluster and its backup storage</title>
+  <desc id="description">The backup cluster uploads SST files to storage; the restore cluster downloads them back when the original cluster is lost.</desc>
+  <defs>
+    <marker id="a" markerWidth="10" markerHeight="10" refX="9" refY="3" orient="auto" markerUnits="strokeWidth">
+      <path d="M0,0 L0,6 L9,3 z" fill="#6f6fff"/>
+    </marker>
+    <marker id="g" markerWidth="10" markerHeight="10" refX="9" refY="3" orient="auto" markerUnits="strokeWidth">
+      <path d="M0,0 L0,6 L9,3 z" fill="#2a9d5c"/>
+    </marker>
+    <marker id="gs" markerWidth="10" markerHeight="10" refX="1" refY="3" orient="auto" markerUnits="strokeWidth">
+      <path d="M9,0 L9,6 L0,3 z" fill="#2a9d5c"/>
+    </marker>
+    <marker id="as" markerWidth="10" markerHeight="10" refX="1" refY="3" orient="auto" markerUnits="strokeWidth">
+      <path d="M9,0 L9,6 L0,3 z" fill="#6f6fff"/>
+    </marker>
+    <marker id="am" markerWidth="10" markerHeight="10" refX="9" refY="3" orient="auto" markerUnits="strokeWidth">
+      <path d="M0,0 L0,6 L9,3 z" fill="#b5760a"/>
+    </marker>
+    <style>
+      .t { font: 600 19px Arial, sans-serif; fill: #170128; }
+      .l { font: 16px Arial, sans-serif; fill: #39314a; }
+      .s { font: 14px Arial, sans-serif; fill: #5b5470; }
+      .m { font: 15px "DejaVu Sans Mono", Menlo, monospace; fill: #39314a; }
+      .mt { font: 600 17px "DejaVu Sans Mono", Menlo, monospace; fill: #170128; }
+      .ms { font: 12px "DejaVu Sans Mono", Menlo, monospace; fill: #ffffff; }
+      .zone { fill: #ffffff; stroke: #b9b3d6; stroke-width: 2; stroke-dasharray: 8 6; }
+      .node { fill: #f3f2ff; stroke: #6f6fff; stroke-width: 2; }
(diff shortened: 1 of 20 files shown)


static/active-active-failover.svg
killme2008, line 1: In an active-active FO solution, cross-region replication is generally absent.

I think we can remove the Region info.

static/BR-explain.svg
killme2008, line 1: The right side should be changed from SST files to backups, which contain not only SST files but also other files.

static/Component-architecture.svg
killme2008, line 1: Nit: Why isn’t object storage as large as the WAL?

killme2008: Or have them drawn like `static/Cross-region-single-cluster.svg`.

killme2008 (approved):
Almost LGTM

killme2008:
Don't forget to backport them into v1.2

static/active-active-failover.svg
fengjiachun: Done in 23e386f — the Region 1 / Region 2 labels and the divider between them are gone, so the diagram no longer implies cross-region replication. Note the overview prose still says "Deploying nodes in different regions can also meet region-level DR requirements, but the scalability is limited", which reads fine as an option rather than the default; say the word if you want that sentence adjusted too.

static/BR-explain.svg
fengjiachun: Done in 23e386f — the six chips now read "Backup file" instead of "SST File".

static/Component-architecture.svg
fengjiachun: That was my mistake, not the original: the current PNG draws the two roughly equal and I made the object storage half the width. Fixed in 23e386f by taking your second suggestion.

fengjiachun: Done in 23e386f — both are now full-width bars stacked under the Datanodes, with a double-headed arrow from each Datanode, the way Cross-region-single-cluster.svg draws them. It also fixes a semantic oddity I had carried over: the old wiring linked the WAL to Datanode 1 and the object storage to Datanode 3, when every Datanode uses both.

fengjiachun:
Extended to 1.2 in c335c86. Version 1.2 referenced the same twelve PNGs and its cross-region page is word for word the nightly one, so that commit only swaps the references — no prose changes, no new files. The PNGs stay in `static/` for 1.1 and older, which still point at them.

fengjiachun:
Resolved the two duplicated region ids in 701512a, then corrected the numbering in 4d61064.

- `Cross-region-single-cluster`: table 2 sits on Datanodes 1, 2, 3 and 5 and now runs `2-1` to `2-4` in Datanode order, so Datanode 3 carries `2-3` and Datanode 5 carries `2-4`. That matches how the other tables are numbered: table 1 runs `1-1` to `1-5` across the Datanodes, table 3 runs `3-1` to `3-4`, table 4 runs `4-1` to `4-2`. My first attempt only renumbered Datanode 3, which restored uniqueness but left table 2 out of order.
- `BR-DR`: cluster 1 holds `Region 1-1` on Datanode 1 and `Region 1-2` on Datanode 2, so the restored Datanode in cluster 2 carries `1-1` and `1-2` instead of `1-2` twice. The ids it shares with cluster 1 are deliberate — they are the same regions, restored from backup.

Correct these in review if the intended distribution was different.

